PR #8, opened 2023-07-31
Remove .godot folder + fix icons

bitbrain:
On 4K monitors, icons looked blurry and tiny - reason being is that we need to enable the 'Scale with editor' in the .svg import settings:

after1
after2

I also removed the `.godot` folder as this one is supposed to be ignored. It contains cached data specific to your machine, so committing this causes other people to get conflicts.


BenjaTK:
Thanks! Probably should've updated .gitignore before the initial commit lol.

PR #13, opened 2023-08-01
Add noise generator

### LINKED ISSUE #2: Add noise-based top-down terrain generation

BenjaTK:
Closes #2 

I made the NoiseGenerator, but I want to see if you find it confusing or not. It works differently from other generators, since instead of one tile, it has a "Tiles" dictionary, which includes thresholds and TileInfo.
image
Basically, for each tile in the world size, it loops through all of these and, if the noise value at those coordinates is higher than the threshold, it places the corresponding TileInfo.
```gdscript
func _set_grid() -> void:
	for x in settings.worldSize.x:
		for y in settings.worldSize.y:
			var noise = settings.noise.get_noise_2d(x, y)
			for threshold in settings.tiles:
				if not (threshold is float) or not (settings.tiles[threshold] is TileInfo):
					continue
				if noise > threshold:
					grid[Vector2(x, y)] = settings.tiles[threshold]
					break
```

It generates results like this:
image

Obviously, it will be a lot better when Godot finishes implementing typed dictionaries. (https://github.com/godotengine/godot/pull/78656)

I'd appreciate the feedback!


BenjaTK:
Btw, this wouldn't work with the GenerateBorders or Walls modifiers, since those get the

Nemesis-AS:
This is as far as it goes with optimization it seems. Though one change I'd recommend is to create another variable that contains only the tiles from `settings.tiles` that match `if not (threshold is float) or not (settings.tiles[threshold] is TileInfo)` and run the loop on that to save a bunch of comparisons, I don't think it would make a ton of difference but it might be worth a shot

BenjaTK:
I could maybe add a setter to the variable so if it finds any values or keys that don't match those types, it deletes it, so you can't even add any.

PR #24, opened 2023-08-06
Improves performance when drawing terrain

BenjaTK:
This significantly decreases the time it takes to draw terrain cells. It's still not perfect, but it works. Here's what the _draw_tiles function takes with a 128x128 `HeightmapGenerator2D` world with the carver modifier, with the new and old code:

image


BenjaTK:
Merging because I don't see any problem and it shouldn't break anything.

Issue #16, opened 2023-08-03
Chunk system?


bitbrain:
I am wondering how much out of scope would be a chunk system - generating an infinite world with this addon (depending on the generator used).

Saving/loading chunks to disk may be out of scope but I could totally see a system where depending on the player position, a level gets re-generated dynamically.

BenjaTK:
That's something I'd like to see! I don't think it would be necessary/make sense with generators like the walker or cellular one (it wouldn't really work), but with noise and heightmap we could add InfiniteNoiseGenerator and InfiniteHeightmapGenerator, and have them be separated from the static ones.
Then the user sets a node as the chunk generation origin, some parameters like `chunkSize` and `chunkGenerationDistance`

BenjaTK:
We also talked about it here #7 as a way of optimizing large world sizes.

titusio:
I'd like to work on something like this. I am not really sure though what would work best. For the sake of compatibility I would use a single TileMap instead of one TileMap node per chunk. The generator could get a separate generate function for a specific chunk and another script could handle the loading management.

BenjaTK:
Closed by #29

PR #29, opened 2023-08-18
Chunk system

titusio:
this is a first version of a chunk system mentioned in #16.
There are some areas which need some work.
- As mentioned in the issue, some generators won't be able to use this system, so introducing a base class for generators using the chunk system could be helpful for differentiating those 2 kinds of generators.
- I think there has to be a solution for saving the chunks. There could be a signal with the grid as a parameter which allows the developer to save the data. A method for adding the data back in could be helpful too.
- In the test scene I added a simple script to load the chunks. Adding a script loading/unloading chunks around a referenced Node2D shouldn't be too hard to add.
- In the test scene I created a unique copy of the Terraria-like settings and disabled the random seed generation. This setting caused problems because the noise on the Caver modifier was different on each chunk. I don't have a solution for this problem.
- This PR currently lacks an unload function.
- Currently only the HeightMapGenerator is implemented.
- There is not documentation for the chunk system.

I hope this PR matches the vision for this plugin :) Any kind of help would be really appreciated ^^


titusio:
I removed some unnecessary code duplication. If we'd introduce a base class for all modifiers which can work with chunk, we could erase this.

BenjaTK:
Just tried it in my editor, looks great! I'm debating rn between having the chunk loader in its own node like it is at the moment, or in the Generators. I think I like the idea of having a `ChunkLoader` node where you set a `radius`, an (optional) `actor`, and maybe an `update_rate` so the chunk loader will automatically load chunks around the actor. What do you think? I don't know if that's the idea you had.

BenjaTK:
Btw, could we add a `has_chunk` function? Could be nice for the chunk loader to avoid regenerating chunks that were already generated.

BenjaTK:
And about the random noise seed problem, maybe we only change the seed in the `apply` function? Instead of the `apply_area` one which is called by both the normal and chunked ones. That _should_ work? Idk how to fix the one for the generator itself, but maybe that's just something the user has to set at the start.

titusio:
I think a `ChunkLoader` node would be best because it makes it clear, that this system is optional. I will add the missing `has_chunk` method together with a better ChunkLoader.
Since the `apply_chunk` method has access to the generator object, I could imagine using a random seed generated by the generator itself. What do you think about introducing a `ChunkAwareModifier` resource (I am really not sure about the name). This way this functionality could be written once and the modifiers would only have to overwrite the `apply_area` method.

BenjaTK:
That sounds good!

titusio:
I think this PR is ready now. Enabling the user to save the data is beyond the scope of this but definitely worth investing some time in. Other than that I think this is finished

addons/gaea/modifiers/carver.gd
```diff
@@ -33,6 +33,22 @@ func apply(grid: Dictionary, _generator: GaeaGenerator) -> Dictionary:
...
+func apply_chunk(grid: Dictionary, generator: GaeaGenerator, chunk_position: Vector2i) -> Dictionary:
+	if random_noise_seed:
+		noise.seed = randi()
+	
+	for x in GaeaGenerator.get_chunk_range(chunk_position.x):
+		for y in GaeaGenerator.get_chunk_range(chunk_position.y):
+			var tile_pos := Vector2(x, y)
+			if not grid.has(tile_pos):
+				continue
+
```
BenjaTK, line 27: Missing
```gdscript
if _is_out_of_bounds(tile_pos):
	continue
```

addons/gaea/modifiers/noise_painter.gd
```diff
@@ -33,6 +33,25 @@ func apply(grid: Dictionary, _generator: GaeaGenerator) -> Dictionary:
...
+	
+	var size: int = GaeaGenerator.CHUNK_SIZE
+	var chunk_origin: Vector2i = size * chunk_position
+	
+	for x in GaeaGenerator.get_chunk_range(chunk_position.x):
+		for y in GaeaGenerator.get_chunk_range(chunk_position.y):
+			var tile_pos := Vector2(x, y)
+			if not grid.has(tile_pos):
+				continue
+
```
BenjaTK, line 29: Same as Carver, missing
```gdscript
if _is_out_of_bounds(tile_pos):
	continue
```

addons/gaea/chunk_loader.gd
```diff
@@ -0,0 +1,113 @@
...
+	
+	# remove old chunks
+	if unload_chunks:
+		var loaded_chunks: Array[Vector2i] = generator.generated_chunks
+		for i in range(loaded_chunks.size() - 1, -1, -1):
+			var loaded: Vector2i = loaded_chunks[i]
+			if not (loaded in required_chunks):
+				generator.unload_chunk(loaded)
+		
+		for required in required_chunks:
```
BenjaTK, line 68: If unload chunks is disabled, it won't generate new chunks. Easily fixed by just getting this for loop out of the if statement.

addons/gaea/generators/generator.gd
```diff
@@ -44,12 +46,28 @@ func generate() -> void:
...
 func erase(clear_tilemap := true) -> void:
 	if clear_tilemap:
 		tile_map.clear()
 	grid.clear()
 
 
+func erase_chunk(chunk_position: Vector2i, clear_tilemap := true) -> void:
+	for x in get_chunk_range(chunk_position.x):
+		for y in get_chunk_range(chunk_position.y):
+			grid.erase(Vector2(x, y))
```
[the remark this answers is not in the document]
titusio: `GaeaGenerator.get_tiles_of_type` is the only method I found that uses Vector2i. I went with Vector2 because all other generators used it.

BenjaTK:
LGTM!

PR #46, opened 2023-09-18
3D support

### LINKED ISSUE #40: HeightmapGenerator3D

### LINKED ISSUE #47: Add 3D support

BenjaTK:
https://github.com/BenjaTK/Gaea/assets/73806216/ccd559bb-fbf3-436b-bf6b-ae2f64ded3a5

Closes #40 and closes #47 

# To-do:
- [x] 3D modifiers
- [x] 3D chunk loading
- [ ] Fix bug with how the mesh renderer looks (see below).
- [ ] NoiseGenerator3D


BenjaTK:
MeshGaeaRenderer has this weird bug where it just looks terrible
image

BenjaTK:
We might want to generalize the classes a bit. Right now, I've added Modifier3D, ChunkAwareModifier3D, GaeaGenerator3D and while it does work I'm not sure if I like the amount of classes there have to be. Also, technically things like the HeightmapPainter could just be one modifier that automatically detects if its a 3D or a 2D generator, which would make things easier.

BenjaTK:
image
Here's what the HeightmapPainter3D looks like!

addons/gaea/chunk_loader_3d.gd
```diff
@@ -0,0 +1,210 @@
...
+						required_chunks.erase(required)
+						break
+			else:
+				generator.generate_chunk(required)
+
+
+func _generate_chunk_threaded(chunk_position: Vector3i, thread: Thread = null) -> void:
+	mutex.lock()
+	generator.generate_chunk.call_deferred(chunk_position)
+	mutex.unlock()
```
chendrak, line 124: I am pretty sure by using `call_deferred`, that method will be executed on the main thread (aka render thread) as opposed to being run in a thread. While I don't use GDScript myself, when Google-ing for `call_deferred`, it was mentioned multiple times that the method will be executed on the main thread.

addons/gaea/chunk_loader_3d.gd
```diff
@@ -0,0 +1,210 @@
...
+	generator.erase.call_deferred()
+	mutex.unlock()
+
+	if thread != null:
+		_thread_complete.call_deferred(thread)
+
+
+func _thread_complete(thread: Thread) -> void:
+	if thread != null:
+		thread.wait_to_finish()
```
chendrak, line 150: I added logs for this one and for me it seems to never be called

chendrak (commented):
**General note:**
This PR introduces 2 different things:
1. 3D support
2. Multithreading

That makes it harder to reason about (as in: is bug X happening because of the multi threading or because of some logic in the 3D support).

**Multithreading Notes:**
From what I can tell, it looks like you are not reaping any benefits from threading since no code is actually executed in a thread (see my comments). Multithreading should probably be builtin to the actual generator, since all generation and modification of things happen in there. That way, you can have a `multithreading` flag on the Generator and it can then always generate using multiple threads or not, no matter if the loading happens chunked or not.

**Notes on the 3D generation:**
1. When thinking about loading chunks in 3D, I would expect `chunk_size` amount of tiles / blocks to be loaded in all directions. It took me a little bit to figure out that it actually generates from `height_offset` all the way down to 0, potentially doing work that is not at all necessary, since the player is not actually close to 0 unless they can dig down. While this doesn't seem to have any noticeable impact on generation time, I find it confusing at best. Generation could still technically stop at 0, but if the player is very high up, the generator shouldn't generate anything outside of `loading_radius`.
2. There is a small bug in the logic when approaching negative coordinates: If `loading_radius` is set to (1,1,1) and you walk left in the demo, the next chunk on the left will only be loaded once you actually cross over into the missing chunk. Moving right works as expected.

I hope all of this makes sense.

addons/gaea/chunk_loader_3d.gd
BenjaTK: I saw that too, but Godot gives me an error when trying to call it normally, and it tells me I should use `call_deferred`. I'm not sure exactly how it works.

BenjaTK:
> **General note:** This PR introduces 2 different things:
> 
>     1. 3D support
> 
>     2. Multithreading
> 
> 
> That makes it harder to reason about (as in: is bug X happening because of the multi threading or because of some logic in the 3D support).
> 
> **Multithreading Notes:** From what I can tell, it looks like you are not reaping any benefits from threading since no code is actually executed in a thread (see my comments). Multithreading should probably be builtin to the actual generator, since all generation and modification of things happen in there. That way, you can have a `multithreading` flag on the Generator and it can then always generate using multiple threads or not, no matter if the loading happens chunked or not.
> 
> **Notes on the 3D generation:**
> 
>     1. When thinking about loading chunks in 3D, I would expect `chunk_size` amount of tiles / blocks to be loaded in all directions. It took me a little bit to figure out that it actually generates from `height_offset` all the way down to 0, potentially doing work that is not at all necessary, since the player is not actually close to 0 unless they can dig down. While this doesn't seem to have any noticeable impact on generation time, I find it confusing at best. Generation could still technically stop at 0, but if the player is very high up, the generator shouldn't generate anything outside of `loading_radius`.
> 
>     2. There is a small bug in the logic when approaching negative coordinates: If `loading_radius` is set to (1,1,1) and you walk left in the demo, the next chunk on the left will only be loaded once you actually cross over into the missing chunk. Moving right works as expected.
> 
> 
> I hope all of this makes sense.

Thanks a lot, I'll look at this when I have time!

addons/gaea/chunk_loader_3d.gd
BenjaTK: Removed multithreading from this PR, no longer relevant.

addons/gaea/chunk_loader_3d.gd
```diff
@@ -0,0 +1,210 @@
...
+## prevents empty areas.
+@export var renderer: GaeaRenderer3D
+## Chunks will be loaded arround this Node.
+## If set to null chunks will be loaded around (0, 0)
+@export var actor: Node3D
+## The distance around the actor which will be loaded.
+## The actual loading area will be this value in all 4 directions.
+@export var loading_radius: Vector3i = Vector3i(2, 2, 2)
+## Amount of frames the loader waits before it checks if new chunks need to be loaded.
+@export_range(0, 10) var update_rate: int = 0
```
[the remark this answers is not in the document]
BenjaTK: Perfect! Done!

BenjaTK:
A look at the HeightmapGenerator3D demo:

https://github.com/BenjaTK/Gaea/assets/73806216/0f14a92f-cf60-4f78-8daf-724b666123e3

addons/gaea/generators/3D/chunk_aware_generator_3d.gd
```diff
@@ -0,0 +1,62 @@
...
+
+func _ready() -> void:
+	if chunk_size == 0:
+		push_error("Chunk Size can not be 0!")
+
+	super._ready()
+
+
+func generate_chunk(chunk_position: Vector3i) -> void:
+	pass
```
titusio, line 25: A little warning message could be useful like `push_warning("generate_chunk method is not overwritten")` .

addons/gaea/renderers/3D/mesh_gaea_renderer.gd
```diff
@@ -0,0 +1,58 @@
...
+@export var flat_normals: bool = false
+
+
+func _draw_area(area: AABB) -> void:
+	if area.size == Vector3.ZERO:
+		mesh_instance.mesh = null
+		return
+
+	var array_mesh: ArrayMesh = ArrayMesh.new()
+	var surface_tool: SurfaceTool = SurfaceTool.new()
```
titusio, line 19: I think constructing the mesh with [ArrayMeshes](https://docs.godotengine.org/en/stable/classes/class_arraymesh.html) would be a lot faster (at least my last project got better performance this way)

addons/gaea/chunk_loader_3d.gd
```diff
@@ -0,0 +1,141 @@
...
+			var loaded: Vector3i = loaded_chunks[i]
+			if not (loaded in required_chunks):
+				generator.unload_chunk(loaded)
+
+
+	# load new chunks
+	for required in required_chunks:
+		if not generator.has_chunk(required):
+			generator.generate_chunk(required)
+
```
titusio, line 88: This doesn't seem to work. Loading other chunks doesn't seem to work too,

https://github.com/BenjaTK/Gaea/assets/80385479/f888eba8-f4f7-4cb7-9a4c-b6080862a47b

BenjaTK: That's because `infinite` isn't true in the HeightmapGeneratorSettings

PR #56, opened 2023-10-22
Add WaveFunctionGenerator2D

BenjaTK:
Uses the WaveFunction2DEntry resource, which holds a TileInfo, a weight and its valid neighbors. It is a bit of work to set up, so we could make a tool or something. But it's generic so it doesn't have to use Tilemaps.

https://github.com/BenjaTK/Gaea/assets/73806216/19e00381-c73c-497c-964f-ec94d70c5192


BenjaTK:
New demo:
https://github.com/BenjaTK/Gaea/assets/73806216/c96736e1-621d-4fe8-88ed-2b77835f28a9

Issue #54, opened 2023-10-19
Print generation time when generation is finished


BenjaTK:

MeBadDev:
It will be great if generation time wont print in a non-debug build

BenjaTK:
Closed with #61

PR #61, opened 2023-10-24
Added generation time debug

MeBadDev:
Closes Issue #54


BenjaTK:
I think it should use miliseconds, with `OS.get_tick_msec`

BenjaTK:
Also, I think if you use `print_debug` it should print with the path to the script and the node, which would be really useful.

BenjaTK:
LGTM!

PR #67, opened 2023-10-26
Make grid a resource

### LINKED ISSUE #65: Make grid a resource
I think the grid could be changed to a GaeaGrid resource, instead of just a dictionary. This resource would still have a dictionary, but it would allow for easier adding of features and saving. 
### The pros:
* Could add `get_value` and `set_value` methods which would allow for doing stuff like #25, #48 or #51.
* As a resource, it can be easily saved and loaded.
* Grid logic would be separated from generation, which is kinda what we've been doing with stuff like separating generation from rendering, and all that.
* Could have helper functions such as getting the neighbors of a tile (which already exist in the Generator class, but is a bit of a mess).

### The cons:
* Lot of refactoring
* Resources can sometimes be a bit wonky
* Can't think of anything more tbh

BenjaTK:
Makes the grid a resource instead of a dictionary as explained in #65 
Closes #65 

This also includes a few changes:
- Vector2 and Vector3 have been changed to their integer variant in the grid.
- Modifier's apply functions no longer return the new grid, they just modify the resource.
- Things like `tile_pos` or `get_tile` have been changed to `cell` for consistency.

TODO:
- [x] Change 3D generators to work with this new resource


addons/gaea/grid.gd
```diff
@@ -0,0 +1,62 @@
...
+extends Resource
+## The grid which all [GaeaGenerator]s fill, and the base
+## of the Gaea plugin.
+
+
+var _grid: Dictionary
+
+
+## Sets the value at [param pos] to [param value].
+func set_value(pos, value: Variant) -> void:
```
chendrak, line 11: Any reason it's not `pos: Vector2i`?

addons/gaea/grid.gd
```diff
@@ -0,0 +1,62 @@
...
+	return _grid.values()
+
+
+## Returns [code]true[/code] if the grid has a cell at [param pos].
+func has_cell(pos) -> bool:
+	return _grid.has(pos)
+
+
+## Removes the cell at [param pos] from the grid.
+func erase(pos) -> void:
```
chendrak, line 40: For this and others that take a Vector2i, maybe you could add convenience methods that just take x and y coordinates?

```
func erase(x, y) -> void:
    erase(Vector2i(x,y))
```

chendrak (approved):

addons/gaea/grid.gd
BenjaTK: This is the generic class and needs to work with both 2D and 3D.

addons/gaea/grid.gd
BenjaTK: Yeah, that sounds good! I'll do that.

addons/gaea/grid.gd
BenjaTK: Btw, I'm pretty sure that with this https://github.com/godotengine/godot-proposals/issues/7450 in 4.2 we'll be able to override the function in those sub classes and set them to what we need!

BenjaTK: OH nvm you can already do this. I did it in GaeaGrid3D without realizing LOL

BenjaTK:
Merging

PR #77, opened 2023-11-03
Warning when attempt to generate from the editor but `editor_preview` isn't enabled.

MeBadDev:
## Should prevent problems like this:
image

## Looks like this:
image


BenjaTK:
Looks great! Thanks!

PR #88, opened 2023-12-25
Changed the way the noise generator accepts data

### LINKED ISSUE #50: Better threshold editor for NoiseGeneratorSettings
Using Dictionaries is just really bad. I'll make a concept soon.

DaviD4Chirino:
I changed from a dictionary to an Array of a custom resource that includes a threshold, TileInfo and a title so one can see what tile its generating.

It skips the sorting.

Also i use [GDFormat](https://www.gdquest.com/tutorial/godot/gdscript/gdscript-formatter/), a godot auto formatter so spacing might have changed

Closes #50


DaviD4Chirino:
Also its my first time adding anithing into another person work, but the way to use the noise generator by making a dictionary with an uncheangeable key it stressed me so...

project.godot
BenjaTK, line 1: This change should be removed as it would probably cause conflicts.

BenjaTK, line 1: This change should be removed as it could cause conflicts.

addons/gaea/generators/2D/noise_generator/noise_generator_data.gd
```diff
@@ -0,0 +1,6 @@
+extends Resource
+class_name NoiseGeneratorData
+##Intended only for organizing
+@export var title: String
```
BenjaTK, line 4: I think the title property should change the `resource_name` of the data. I did the same for TileInfo I think.

addons/gaea/generators/2D/noise_generator/noise_generator_data.gd
```diff
@@ -0,0 +1,6 @@
+extends Resource
+class_name NoiseGeneratorData
+##Intended only for organizing
```
BenjaTK, line 3: Also, just for consistency, please add 2 empty lines after the class_name and reverse the order of the 2 first lines like the other scripts. Also adding a small description to the class and variables would be nice.

scenes/demos/noise/noise_demo.tscn
BenjaTK, line 1: Updating this demo to match the other one would also be nice, shouldn't be hard.

BenjaTK (changes requested):
This looks great! Well done! This will definitely make it easier. Another thing we could add is making the threshold have a max and the minimum (either as the default or as an optional change).

BenjaTK:
Also I'm so sorry for being so late.

project.godot
DaviD4Chirino: I tried to downgrade it into godot 4.1 but it uses syntax from 4.2, so i assumed you mean the gl_compatibility so i changed that, let me know if the version causes problems.

addons/gaea/generators/2D/noise_generator/noise_generator_data.gd
```diff
@@ -0,0 +1,15 @@
...
+extends Resource
+## Data for [NoiseGenerator]s, it contains a [param threshold] and a [param tile] of the class [TileInfo], also a [param title] for easier identification.
+
+## [param title] will rename the resource when it changes
+@export var title: String = "":
+	set(value):
+		title = value
+		resource_name = title
+## [param threshold] if the threshold is higher than 
+## the noise it will place the [param tile] there
```
BenjaTK, line 12: I think this could be changed to "If the noise value at a cell is lower than [param threshold], [param tile] will be placed.

addons/gaea/generators/2D/noise_generator/noise_generator_data.gd
```diff
@@ -0,0 +1,15 @@
...
+@export_range(-1.0, 1.0) var threshold: float
+## The tile to place
+@export var tile: TileInfo = TileInfo.new()
```
BenjaTK, line 15: The default value shouldn't be TileInfo.new(), it should be null.

BenjaTK (commented):
Everything else LGTM!

addons/gaea/generators/2D/noise_generator/noise_generator_data.gd
DaviD4Chirino: i just implemented the ranged thresholds and im wrapping up, i have a group named thresholds with a min and max range, how about you check this first? 
```
## To allow for a range of thresholds
## Also min and max have a setter that prevents min from being greater than max and vice versa
@export_group("Thresholds")

## The minimum threshold
@export_range(-1.0, 1.0) var min: float = -1.0 : 
	set(value):
		min = value
		if min > max:
			max = min

## The maximum threshold
@export_range(-1.0, 1.0) var max: float = 1.0 : 
	set(value):
		max = value
		if max < min:
			min = max
```

BenjaTK: Looks great! Do groups have tooltips? I don't think they do.

BenjaTK: I would also recommend renaming it to threshold_min and threshold_max and doing @export_group("Thresholds", "threshold_"), just for clarity.

DaviD4Chirino: Did not know about that, actually

BenjaTK (approved):
LGTM! Just need to solve those conflicts
image

BenjaTK:
Btw, is there any way you could remove the files that don't need to be changed? Like project.godot, so only the files you changed stay.

DaviD4Chirino:
Yeah it should be posible, in this case project.godot is mostly static
except for the version

BenjaTK:
This all looka great! Thank you so much! Merging rn.

DaviD4Chirino:
yeah its ready, also i saw your preview noise test and i tink i i can implement it, so when that is ready i will modify acordingly

pd: i only saw the first part of the last message, so thats why i took so long to answer

BenjaTK:
Thanks a lot for your contribution!

PR #95, opened 2024-01-15
Fix conversion from ms to s

nubels:
The conversion from milliseconds to seconds was wrong. It needs to be divided by 1000, not 100.


BenjaTK:
Thanks a lot!

BenjaTK:
It seems all the generators have this mistake, could you check the other ones, please?

nubels:
> It seems all the generators have this mistake, could you check the other ones, please?

Sure, looks like I've missed two. That should be all now.

PR #97, opened 2024-02-05
Fixed NoisePainter3D. Replaced `grid.set_cell` with `grid.set_value`

RAMENtheNOODLES:
Let me know if this should not be changed. I found this while using Gaea for my own game, and it was bringing up errors when using noise_painter_3d, but when I switched the function call, it worked like normal.


BenjaTK:
oh, how did I miss this??? thanks!

PR #111, opened 2024-05-06
Add `enabled` setting to Modifier

nedmas:
I find it really useful to be able to turn modifiers on/off while I'm building and debugging generated content.


BenjaTK:
This is a really good idea, thanks!

PR #116, opened 2024-05-30
Add threaded chunk loaders and renderers

cullumi:
Implemented some threaded chunk loaders (wrappers for the non-threaded ones). Mostly just intended to get the bulk of the loading process off of the main thread.


cullumi:
There's still some stuttering that happens, but I haven't quite narrowed down the cause.

addons/gaea/generators/2D/chunk_aware_generator_2d.gd
```diff
@@ -36,8 +36,8 @@ func erase_chunk(chunk_position: Vector2i) -> void:
 			for layer in grid.get_layer_count():
 				grid.erase(Vector2i(x, y), layer)
 
-	chunk_updated.emit(chunk_position)
-	chunk_erased.emit(chunk_position)
+	(func(): chunk_updated.emit(chunk_position)).call_deferred() # deferred for threadability
+	(func(): chunk_erased.emit(chunk_position)).call_deferred() # deferred for threadability
```
BenjaTK, line 40: I think you can just do something like `chunk_erased.emit.call_deferred(chunk_position)`, instead of having to create a lambda. Would work the same probably just easier to read.

cullumi: I thought the same thing, but surprisingly enough, you actually can't. The reason why is that Signals aren't Objects, so therefore methods called on them cannot be made into Callables. I found the recommendation of using lambdas in the Callable documentation iirc.

cullumi:
> There's still some stuttering that happens, but I haven't quite narrowed down the cause. Pretty sure lag spikes are much shorter in length than when I started, though.

I figured out that the stuttering came from the Tilemap and Gridmap GaeaRenderers. I threaded them too and there is hardly any lag now. You do technically run a greater risk of falling into pits and the like, though there are ways around that I'm pretty sure.

The rate at which it renders can be improved by running each render task in parallel (currently it waits for one to complete, then moves on to the next in a queue).

cullumi:
Made is so that the Threaded Gridmap and Tilemap renderers can render chunks in parallel.

Next, I need to make them gracefully handle "misses" that I suspect are resulting from trying to un-render chunks that have already been or are being un-rendered.

cullumi:
Fixed the errors I was encountering. It wasn't a logical error ("misses") after all, it was just that there were still a few method calls to the Tilemap and Gridmap that needed to be deferred. The error is gone now and the crashes that they were causes have stopped as well.

Tbh this is pretty much ready to go at this point.

addons/gaea/renderers/2D/threaded_tilemap_gaea_renderer.gd
```diff
@@ -0,0 +1,36 @@
+@tool
+class_name ThreadedTilemapGaeaRenderer
+extends TilemapGaeaRenderer
+## @experimental
+
+@export var threaded:bool = true
+@export var max_running:int = 1
```
BenjaTK, line 7: Could you add a description to this that says that setting it to -1 will uncap it? And maybe make it an `export_range` with a minimum of -1 and an uncapped max.

addons/gaea/renderers/3D/threaded_gridmap_gaea_renderer.gd
```diff
@@ -0,0 +1,36 @@
+@tool
+class_name ThreadedGridmapGaeaRenderer
+extends GridmapGaeaRenderer
+## @experimental
+
+@export var threaded:bool = true
+@export var max_running:int = 1
```
BenjaTK, line 7: Same as the Tilemap one.

addons/gaea/others/threaded_chunk_loader_2d.gd
BenjaTK, line 1: And would it be possible to have multiple tasks in ChunkLoaders like how it is in the renderers? You can still see the chunks loading especially in 3D.

BenjaTK (changes requested):
Ngl this brings tears to my eyes, it's beautiful. 3D is so much smoother now. This could actually push us to the 1.0 release and to upload it to the Godot Asset Library.

addons/gaea/others/threaded_chunk_loader_2d.gd
cullumi: The answer is yes! But also it'll take digging into the details a bit more and breaking up the original _update_loading function in a way that can be parallelized. Sooo... probably best saved for a separate PR tbh.

BenjaTK (approved):

Issue #117, opened 2024-06-05
ChunkLoader3D causes "Out of bounds get index '0'" error when "load_on_ready" is set to true and generator's "infinite" setting is set to false.


cullumi:
The simple solution ( implemented in PR #118 ) is copying the following warning/early return from it's _process method up to _ready:
```
if generator.settings.get("infinite") == false:
	push_warning("Generator's settings at %s has infinite disabled, can't generate chunks." % generator.name)
	return
```

However, this error doesn't come up for the very similar ChunkLoader2D, so I'm making an issue in case the real issue lies in the HeightmapGenerator3D's initialization.

BenjaTK:
Ok, it seems to be a problem with the `has_cell` method. It was checking a layer that didn't exist, so just adding a `grid.has(layer)` check fixes it.

BenjaTK:
I've also removed the check for `infinite` enabled since you should be able to use ChunkLoaders on finite worlds.

PR #118, opened 2024-06-05
Chunk loader non infinite error fix

cullumi:
See issue #117 

Additionally this pull request contains an option for enabling or disabling the process method on the ChunkLoader3D, for the situation where one would like to disable but not delete the Chunk Loader (stops all the warnings if having infinite = true is intentional).


BenjaTK:
Could you also add `enabled` to ChunkLoader2D? It does sound useful.

BenjaTK:
Also, ChunkLoaders should be able to generate chunks for non-infinite worlds. In that case they'd be used to optimize it.

BenjaTK:
If you could change this PR to just be about adding `enabled` to both ChunkLoaders I'd appreciate it.

PR #123, opened 2024-06-06
Fixed issue with mismatching tile sizes and added utility functions

BenjaTK:
The renderers will automatically make the tile sizes match when the TileMap's tile size/GridMap's cell size doesn't match with the generator's tile size, which causes visual offset


BenjaTK:
I could either make it so it's set automatically or push an error so the user changes it manually. Not sure which one is better.

PR #140, opened 2024-07-17
Add serialization functions

BenjaTK:
Adds 2 functions to GaeaGenerator:
- `serialize` returns a `PackedByteArray` that represents the currently generated grid.
- `deserialize` takes these bytes and sends them to the grid.

Is this enough for serialization or should we also add functions to save for files? I feel like that should be left to the user, but if there's enough demand I'll add them.

### DIFF
--- a/addons/gaea/generators/generator.gd
+++ b/addons/gaea/generators/generator.gd
@@ -73,7 +73,7 @@ func get_seed() -> int:
 	return seed
 
 
-## Returns the grid as a [PackedByteArray], allowing you to deserialize it later using [method deserialize].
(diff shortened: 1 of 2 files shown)


cullumi:
> Is this enough for serialization or should we also add functions to save for files?

I think it's enough for this PR at least, since saving to a file is very context-specific.

But a lightweight implementation for demo purposes might not hurt. Anything beyond that may very well baloon out of the scope of what makes sense for a Toolkit, though.

BenjaTK:
That's a good point, I'm going to merge this.

PR #144, opened 2024-07-25
Add Advanced Modifier

BenjaTK:
AdvancedModifier(2D/3D) is a modular modifier which will allow for much more complex generation. 

It works with an array of **conditions**, all of which it have to be met to place the tile on the target cell. 

For example:
Let's say we want to place vegetation on top of side-scrolling terrain. We'll use an OffsetCondition, which checks for a certain tile in an offset to allow placing.
image

In this case, the first condition is only met if there is a ground tile below.
We also have a chance condition, which is pretty self explanatory.

image

Now we'll make the tile to be placed a RandomTileInfo, which will choose between trees, grass and tall grass.

Here's the result:
image

We can also do stuff like this (the torches):
image

This PR also adds a filter type to only apply modifiers on empty tiles.

---

## To-do
- [x] OffsetRule3D
image
- [x] Documentation
- [x] Icons for the modifier and rules


BenjaTK:
I am also not entirely sure about the wording of "rules"? Any ideas?

PR #158, opened 2024-08-15
Update to 4.3 (Support TilemapLayers)

### LINKED ISSUE #136: Update to Godot 4.3 (TileMapLayer)
I know 4.3 isn't out yet but is there currently a way I can assign a TileMapLayer node to a generator rather than a Tilemap node? Couldn't find any info about this

BenjaTK:
Closes #136 
This PR allows using both TileMapLayers and TileMaps, though it makes TileMapLayers the default.


BenjaTK:
Gonna let this one sit for a while so folks can test it out. Also walker_demo.tscn is broken and I can't tell if that's a 4.3 thing or it already happened and I didn't notice.

addons/gaea/renderers/2D/tilemap_gaea_renderer.gd
adadgio, line 1: When using tilemap layers instead of tile map as shown in the screenshot, the _ready function of tilemap_gaea_renderer will produce an error.

Screenshot_2024-08-18_at_08_54_55

adadgio: The following fixes the issue: 

```
func _ready() -> void:
	super()
        **## CHANGE
	if tile_map:**
		if Vector2i(Vector2(tile_map.tile_set.tile_size) * tile_map.scale) != generator.tile_size:
			push_warning("TileMap's tile size doesn't match with generator's tile size, can cause generation issues.
						The generator's tile size has been set to the TileMap's tile size.")
			generator.tile_size = Vector2(tile_map.tile_set.tile_size) * tile_map.scale

```

BenjaTK: Fixed in latest commit!

PR #163, opened 2024-08-25
Produce error if GaeaRenderer doesn't have a generator assigned

adadgio:
When the tilemap rendere does not have a generator assigned, this produces a low level error. Its best to warn users explicitely, as per usual, of the missing implementation.

### DIFF
--- a/addons/gaea/renderers/2D/tilemap_gaea_renderer.gd
+++ b/addons/gaea/renderers/2D/tilemap_gaea_renderer.gd
@@ -32,6 +32,12 @@ enum NodeType {
 
 func _ready() -> void:
 	super()
+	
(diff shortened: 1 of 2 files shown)


BenjaTK:
Could you please remove the other commits unrelated to the PR?

BenjaTK:
Also this could be added to the GridmapRenderer too, and then to the script templates for renderers added in the project.

adadgio:
> Also this could be added to the GridmapRenderer too, and then to the script templates for renderers added in the project.

I don't really know how to do that unless my PR merges into main.

BenjaTK:
Thanks for contributing!

adadgio:
@BenjaTK Your addon is awesome, it was a meagre contribution, but a pleasure.

PR #219, opened 2025-03-30
Make RandomMaterial use weights for better modularity

Raftatul:
This feature gives more control over which tile should be spawned when using the ran_material resource.
The weight element is automatically created when a material is added.
The higher the value, the greater the chance of using the material.

https://github.com/user-attachments/assets/d22d67c4-73f8-41b2-b0e4-6f15beb00a56

### DIFF
--- a/addons/gaea/resources/materials/random_material.gd
+++ b/addons/gaea/resources/materials/random_material.gd
@@ -13,7 +13,7 @@ extends GaeaMaterial
 
 
 ##Represent the weight of each material.[br]
-##For example, if you want to change the probability of spawning the first material, you need to modify the value of the first element in the weight array.[br]
+##For example, if you want to change the probability of using the first material, you need to modify the value of the first element in the weight array.[br]
 ##Higher values increase the chances of obtaining the material.[br]
 ##The object picked is determined by the [method RandomNumberGenerator.rand_weighted] function, [url=https://docs.godotengine.org/en/latest/tutorials/math/random_number_generation.html#weighted-random-probability]check the documentation here.[/url]
 @export var weights: PackedFloat32Array : 
--- a/addons/gaea/resources/materials/random_material.gd
+++ b/addons/gaea/resources/materials/random_material.gd
@@ -7,7 +7,7 @@ extends GaeaMaterial
 	set(value):
 		materials = value
 		weights.resize(materials.size())
-		weights[-1] = 1.0
+		weights[-1] = 100
 		
 		notify_property_list_changed()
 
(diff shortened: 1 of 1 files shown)


cullumi:
Works as advertised! You even made it so the user can't make the weights array a different length than he materials array.

Raftatul:
Yeah, the user can still rearrange the order of weights and tweak the values without breaking anything.

addons/gaea/resources/materials/random_material.gd
```diff
@@ -3,8 +3,27 @@ class_name RandomMaterial
 extends GaeaMaterial
 
 
-@export var materials: Array[GaeaMaterial]
+@export var materials: Array[GaeaMaterial] :
+	set(value):
+		materials = value
+		weights.resize(materials.size())
+		weights[-1] = 1.0
```
BenjaTK: Instead of doing `[-1]`, I think it's nicer to do `set(-1,1.0)`.

addons/gaea/resources/materials/random_material.gd
```diff
@@ -3,8 +3,27 @@ class_name RandomMaterial
...
+	#Avoid editing the weights array
+	set(value):
+		if value.size() == materials.size():
+			weights = value
+
+
+##Return the random picked material. 
 func get_resource() -> GaeaMaterial:
-	return materials.pick_random().get_resource()
+	var random = RandomNumberGenerator.new()
```
BenjaTK: This should have it's seed set to the generator's seed.

BenjaTK (changes requested):
I think weights should be ints instead. That's how we do it in the nodes, at least. 1.0 would be 100, for example.

addons/gaea/resources/materials/random_material.gd
Raftatul: I changed the method to get the material. Now, it should always return the same material when using the same seed in the generator.

BenjaTK: Where's the seeding?

Raftatul: The seed is currently set in floor_walker, but it should be set directly in the material or before starting the loop that calls every get_resource(), so the seed isn't set every time get_resource is called.

BenjaTK: Hmm, it _should_ work fine since nodes with randomization should already set a seed anyways. Otherwise we can fix it later.

BenjaTK (approved):
LGTM

Issue #218, opened 2025-03-30
[2.0] Color Readability for Slot Types


Zehir:
### Problem
Hello,

I have difficulty distinguishing the colors used for slot types. With my eyes, I see one red and five shades of grey, making it hard to differentiate between them.

### Suggestion
To improve readability, I suggest using either distinct shapes or more distinguishable colors. When selecting colors, I usually refer to [this palette](https://flatuicolors.com/palette/defo) from [Flat UI Colors](https://flatuicolors.com/palette/defo).

### Recommended Colors
Here are some colors that are visually distinct:
- **Emerald**
- **Peter River**
- **Amethyst**
- **Wet Asphalt** (maybe not, as it’s quite dark and could be invisible on dark background)
- **Sun Flower**
- **Alizarin**
- **Clouds**

These colors are far apart in contrast, making them easier to distinguish.

BenjaTK:
Sadly Godot doesn't allow to use custom shapes for this I think, but I appreciate the feedback! Will look into changing the colors.

Raftatul:
You mean something like this ?

![Image](https://github.com/user-attachments/assets/ce983c57-1382-48d4-9501-55c684434861)

PR #220, opened 2025-03-31
Improve display for Output Node with name and state

Zehir:
# Improve display for Output Node with name and state

## Changes Introduced:
- Improved alignment of text and slot ports for better visual consistency.
- Displayed the `resource_name` of the `GaeaLayer` on the **Output** node.
- Indicated when the `GaeaLayer` is **disabled**, making its status clearer.
- Updated the slot's left and right labels to use `RichTextLabel` with BBCode.

These changes enhance usability and clarity when working with GaeaLayer nodes.

Before :
image
After :
image

image


Zehir:
@Raftatul you may want to checkout this PR, I just seen your response here : 
https://github.com/BenjaTK/Gaea/issues/218#issuecomment-2766452470

BenjaTK:
Oh this slaps. I don't think it needs the `(Disabled)` label, just darkening the text should work fine? You have to disable it manually anyways so you'd know what caused it

Raftatul:
> @Raftatul you may want to checkout this PR, I just seen your response here : [#218 (comment)](https://github.com/BenjaTK/Gaea/issues/218#issuecomment-2766452470)

Yeah, I'm only changing the slot icons on the node graph, so it shouldn't break anything. But I will still have the issue of some icons not using the right one.

Zehir:
I tried with strike text and a prefix with the layer index;

image

We could also add a tooltip but it will be shown instead of the default one with the output description.

Zehir:
I just notice that the bullet offset it because of the editor scale.

Before my PR : 
image

After the PR :
image

So I will need to revert the alignement fix and find from where it's came from. Maybe it's a godot issue because the dot are smaller too with a bigger editor scale

BenjaTK:
Does this update when the layer order is changed?

Zehir:
Ok now it does @BenjaTK

BenjaTK:
Could you make it so it ignores adding null layers, please?
image
Like the `<empty>` one here.

BenjaTK:
Or adds it as a disabled layer, I guess.

Zehir:
Yes, that could work — an empty entry could be displayed in red to indicate it's invalid.  
Not displaying it would mess up the ID.

We could also add a button to create one, and a pen icon to edit the name using a small text input.

Another idea would be to have an `Array[GaeaLayer]` in the generator, and a `Dictionary[GaeaLayer, TileMapLayer]` in the renderer, so you never need to rely on the layer index.
image

However, this would require saving the layer as a resource somewhere in your project. If both Generator and renderer are in the same scene tree file you can reference them with a copy/past without saving them in a separate file.

Also worth noting: you won't be able to use the same layer for multiple TileMaps in the renderer — but is that really a problem?

BenjaTK:
I think it's fine having the renderer thing be an array, because of what you said about the referencing. Godot can also be a bit finicky about references.

Zehir:
what about that ?

image

BenjaTK:
I think the width increase is fine.

Zehir:
Also the slot could instead of the simple text an text input for the name and a toggle at the end to enable/disable the layer.
And you don't really need to expose the resource in the inspector except if you plan to add more things.

And in the renderer it's auto add the layer instead of an array like with the parameters

BenjaTK:
And iirc the layer won't generate without a GaeaLayer resource

Zehir:
Updated PR with auto resize in X and Y. Also add a warning if the GaeaLayer is missing

Zehir:
About the idea of having a way to edit the name and the enabled state, should I add it ? Could be done this weekend

BenjaTK:
I think that should be its own PR later on

Issue #222, opened 2025-03-31
Graph Node - Slot icon issue


Raftatul:
## Slot icon issue

> You can set custom slot icons on graph nodes using the function set_slot or set_slot_custom_icon_left / set_slot_custom_icon_right.
> I'm working on a PR to add icons to specific types. It didn't work for every node, though, and I don't understand why.
> 
> ![Image](https://github.com/user-attachments/assets/6eda02a2-34a3-42bd-9faf-ad6b09edae1f) 

 _Originally posted by @Raftatul in [#218](https://github.com/BenjaTK/Gaea/issues/218#issuecomment-2766452470)_

Hi, I'm working on a PR to add custom slot icons on graph nodes based on their type. Overall, it's working, but some slots are not detected as the correct type.

For example, the Vector3 input 'To Point' on the 'DistanceFilter' node should have the Gaea icon, but it's still showing the default circle.

Same issue with input 'Amount' of type int of the node 'RandomScatter' not showing the triangle icon. 

I'll open the PR so you can read the code.

![Image](https://github.com/user-attachments/assets/6b1523c9-ff1f-499b-a75f-ef2a0006d23c)

PR #224, opened 2025-03-31
Add separate icons for each slot type

Raftatul:
Made a mistake in the previous PR
This PR allows to fix one of the issues raised in #218
But some things still need to be fixed. I talk more about it in #222


Zehir:
Could we use the built-in icons for known types like vectors ?

Raftatul:
Yeah, but for that we need make the function not static. We can use the function get_editor_theme to get the built-in icons but it can't be called from a static function.

BenjaTK:
> Could we use the built-in icons for known types like vectors ?

I don't think we should, imo they should just stay shapes. VisualShaders in Godot just use circles, but I agree with the idea of changing them to different shapes.

Raftatul:
Yeah, it's not very visible
image

BenjaTK:
Btw, just as a bit of a nitpick. I don't really like the star symbol. Maybe just a square? For other shapes we could have hexagons, for example

BenjaTK:
Also, remember to optimize the icons with svgo! They have a web interface, and Godot has a docs page on how you should make icons for it.

Zehir:
For reference svgo is here : https://jakearchibald.github.io/svgomg/

Zehir:
Hello, I made a [PR](https://github.com/Raftatul/Gaea/pull/1) to help you with #222

Long story short, it's was an issue with how the slot was trying to figure out his ID in the GaeaGraphNode. I tried multiple times to make it work but each time something else was messed up.

I also updated the icons and colors, here the result :
image
image

About the miss alignement of the icons it's a Godot issue : https://github.com/godotengine/godot/issues/98513

BenjaTK:
Superseded by #245

PR #226, opened 2025-04-01
[2.0] Chunk Loader Fixes

cullumi:
## Purpose
There were a number of issues with the Chunk Loader(s) limiting the functionality of infinite world generation.

This PR aims to do the following:
- Get the ChunkLoader2D initialization.
  - It no longer fails to initialize when the actor starts within the origin chunk (0, 0).
- Fix the Chunk Loader tiling the origin chunk for all chunks.
  - This was not an issue for simplex noise, as it is already based on world position.
  - Solution is to add salt to the seed of any randomization graph nodes based on AABB position.
- ~~Add an offset field for use in the generator's noise and randomization graph nodes.~~
  - ~~This makes it possible to, for example, center notable features on the origin.~~
  - ~~This offset is also added to the salt for randomization graph nodes.~~

## Notable Features
- A Simple 2D Character for use as an actor for Chunk Loaders
- Properly functioning Chunk Loader 2D
- A Chunk Loader 2D demo scene
- Better position-based generation for Gaea Generator's various graph nodes.
- ~~An AABBO data structure; it simply contains an AABB paired with an offset.~~
  - ~~This could probably be done better, but it also allows us to add parameters and things for the graph nodes more easily in the future (just adding the 'O' to every single one was painful as is; adding actual function parameters would be worse).~~


BenjaTK:
Why AABBO and not just ofsetting the position of AABB?

cullumi:
> Why AABBO and not just offsetting the position of AABB?

Tbh, I'm looking back on this and feeling that the whole way I went about it was a bad idea; it'd be much better handled as a dedicated graph node, or a graph parameter directly hooked up to the particular noise / generator graph nodes.

I was definitely feeling iffy about it to begin with, which I suppose why I left this as a draft.

BenjaTK:
This is becoming a bit stale and has a lot of conflicts, I'd recommend a remake

PR #231, opened 2025-04-04
Allow making the Gaea editor floating

### LINKED ISSUE #200: Add ability to pop out the panel

BenjaTK:
Closes #200 

Peek 2025-04-03 22-26


cullumi:
Working great on my dual-monitor setup whichever screen I popped it out from! Though, I found my first instinct was to assume that clicking the pop out button again was how to re-dock it, so that would probably make it feel more complete.

BenjaTK:
I based the fact that it doesn't re-dock it on how Godot does it, which I guess makes sense since the button itself looks like a window popping out, but also I understand what you mean.

cullumi:
I suppose an alternative would be to hide the button entirely when it's in window mode? Then at least the button won't even exist to mislead me 😆

BenjaTK:
@cullumi ok, I changed it.

PR #235, opened 2025-04-05
Improve `weights` Array Handling in `RandomMaterial`

Zehir:
# 🛠️ Refactor: Improve `weights` Array Handling in `RandomMaterial`

## Summary

This PR refactors the management of the `weights` array in the `RandomMaterial` resource to ensure correct initialization and consistent resizing behavior.

### Problem

Previously, only the last element in the `weights` array was initialized with the default value of `100`. This could lead to unintentional bias or incorrect behavior when generating random materials.

### Solution

- Improved the initialization logic to ensure that all weights are properly set to a default value.
- Refactored the resizing behavior to maintain consistency across the entire array.
- Initialized `weights` values in the `floor_walker.tres` resource files for data consistency.

### Notes

> We could eventually switch to using `rand_weighted()` from Godot's `RandomNumberGenerator` for more robust random selection.  
> This would require converting the `weights` `PackedInt32Array` to a `PackedFloat32Array`:  
> 📚 [Godot Documentation – rand_weighted](https://docs.godotengine.org/en/latest/classes/class_randomnumbergenerator.html#class-randomnumbergenerator-method-rand-weighted)

---

Let me know if you prefer to switch to `rand_weighted()` now or keep the current manual handling.

### DIFF
--- a/addons/gaea/resources/materials/random_material.gd
+++ b/addons/gaea/resources/materials/random_material.gd
@@ -22,7 +22,7 @@ var _random = RandomNumberGenerator.new()
 	set(value):
 		var pre_size: int = materials.size()
 		materials = value
-		
+
 		#Resize the weights array to match the materials array size
 		if materials.size() > pre_size:
 			weights.resize(value.size())
@@ -30,9 +30,8 @@ var _random = RandomNumberGenerator.new()
 				weights[index] = 100.0
 		elif materials.size() < pre_size:
 			weights.resize(value.size())
-		
-		notify_property_list_changed()
 
+		notify_property_list_changed()
 
 ## Represent the weight of each material.[br]
 ## For example, if you want to change the probability of using the first material, you need to modify the value of the first element in the weight array.[br]
@@ -46,10 +45,10 @@ var _random = RandomNumberGenerator.new()
 			weights = value
 
 
-## Return the random picked material. 
+## Return the random picked material.
 func get_resource() -> GaeaMaterial:
 	var material_index: int = _random.rand_weighted(weights)
-	
+
 	if material_index != -1:
 		return materials[material_index]
 	return null
--- a/addons/gaea/resources/materials/random_material.gd
+++ b/addons/gaea/resources/materials/random_material.gd
@@ -11,7 +11,8 @@ extends GaeaMaterial
 		#Resize the weights array to match the materials array size
 		if materials.size() > pre_size:
 			weights.resize(value.size())
-			weights[-1] = 100
+			for index in range(pre_size, materials.size()):
+				weights[index] = 100
 		elif materials.size() < pre_size:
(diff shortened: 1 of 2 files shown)


Zehir:
Also about the Float switch, for an other upcoming PR I will use the `Probability` value of the tile from the tile set and they are float so you will make my life easier if they are float instead.

https://docs.godotengine.org/en/stable/classes/class_tiledata.html#class-tiledata-property-probability

Raftatul:
I agree with using a float for the probability (so we can use rand_weighted).
At first, I was using a custom array for the weights because I thought using the probability directly from the tileset would be annoying if you wanted to change it for another random material.
But since we can create tile variants and modify their probabilities, I guess that works fine.

Although, creating a tile variant seems to reset every attribute of the tile

Zehir:
I think we should keep the weight array here because you may want to have it in the tree of random material when you want to have a random type like mountain or grass and then a specific mountain random.

But using only the probability would be a mess to setup. I think they need to be keeps like this. But with float instead of int.

My idea is to use the tile set probability to generate the random material resource but that's is for an other PR

BenjaTK:
I wouldn't mind switching it to floats, I just said that in the original PR because that's what the SnakePath2D node does (and I still think it should). maybe still keeping it as 100 for the default? just gives more room for tweaking without going into multiple decimal places.

Zehir:
Like @cullumi suggested I created an seperate issue for the other idea ; #236

The idea is to use what is already existing and avoid duplicate code.

No problem, I will update the PR to use float with default value of 100.0

Zehir:
Ready for review

addons/gaea/resources/materials/random_material.gd
```diff
@@ -12,41 +27,28 @@ extends GaeaMaterial
...
-			return i
-	
-	return -1
-
-
-##Return the random picked material. 
+## Return the random picked material. 
 func get_resource() -> GaeaMaterial:
-	var material_index: int = get_random_material_index()
+	var material_index: int = _random.rand_weighted(weights)
```
BenjaTK, line 50: Has this been tested with the same seed? Don't think it'll be seeded.

Zehir: It's will give different result with the same seed like before.
Do you want to make it generate the same result with the same seed ?

BenjaTK: Yup. The only problem is we can't set the seed in `get_resource` because it'll use the same material every time. I have an idea on how to fix this except it'd be its own PR

Zehir: Yes it's would need some kind of access to `generator_data` like in `snake_path_2d` to allow other use cases too or just a `RandomNumberGenerator` instance for that case.

PR #240, opened 2025-04-05
Add Falloff node

BenjaTK:
image

This still needs another PR in work that adds a node to multiply 2 data inputs together, and then can be used with noise to generate islands. Not sure if it'll work with 3D?


Zehir:
For a 3D I suppose you would get sphere like shape or in this case a cube with smooth edges.
More or less like the DistanceFilter node but with a gradient instead of a filter

BenjaTK:
For 3D I think you could actually use a node that takes the 2D noise data and transforms that to height in the cells, which I think I'll try to make at some point.

Issue #247, opened 2025-04-06
Add Reroute node


Zehir:
**Is your feature request related to a problem? Please describe.**
When the graph became quite complete it's starting to be a challenge to follow wires when they are behind nodes

**Describe the solution you'd like**
I would like to add the Reroute node like in the shader graph

![Image](https://github.com/user-attachments/assets/59205f4e-41b1-42af-848f-beaf50ba2aa1)

Zehir:
Maybe this should be worked on after #221 because it's will also need an input limit of 1 on each reroute node

PR #249, opened 2025-04-06
Add context menu for link interactions

Zehir:
Introduce a context menu for managing link connections within the graph editor, allowing users to disconnect nodes directly via the right-click menu.

This PR is a kind of low-hanging fruit that will be useful for #247.

image
image
You don't see it but the mouse is on the top left of the popup box

You're most likely selecting the correct link with a right-click — even when doing something silly like clicking on a link intersection.

I kept the behavior consistent with the Shader Editor: links take priority.


Zehir:
Note sure why godot keep changing the image sub resource but each time I save they get updated. I don't even know what image they are

Issue #230, opened 2025-04-03
Run the generation on its own thread


BenjaTK:
**Is your feature request related to a problem? Please describe.**
Generation can be a bit slow with more complex graphs, so a performance boost is needed.

**Describe the solution you'd like**
We could either:
- Run the whole generation on a different thread than the main thread, avoiding the game being stopped.
- Run each layer's generation on different threads.

The first option would be the simpler one, but the second option could improve performance even more.

cullumi:
The first will be quite easy to achieve overall, the second (multithreading) will need some further thought.

Some potential sub-goals / topics:
- a node's data should only be calculated once, if possible. This probably necessitates saving the results somewhere accessible from each thread.
- non-thread safe code needs to be accounted for anywhere it shows up. (thread_safe() calls wrapping such code is usually enough for this). This will continue to come up as more nodes are added.

BenjaTK:
Do you think it's a good idea to stop adding nodes until adding thread-safety, since it could be a big refactor afterwards?

cullumi:
> Do you think it's a good idea to stop adding nodes until adding thread-safety, since it could be a big refactor afterwards?

I think as long as that refactor is part of the PR it won't really be an issue. We could also add a little "thread safe" boolean to node resources to mark them as safe or not; at which point we could run ones that aren't set up for it in a fully threadsafe way.

cullumi:
Oh and here's the documentation on thread safe APIs:

[Godot 4.4 - Thread Safe APIs](https://docs.godotengine.org/en/stable/tutorials/performance/thread_safe_apis.html)

BenjaTK:
@cullumi what do you think on getting started with at least the first solution?

cullumi:
> @cullumi what do you think on getting started with at least the first solution?

Can do! I think I *started* working on this a while back, I can pick it up again.

PR #252, opened 2025-04-07
Generation Caching and Traversal

### LINKED ISSUE #233: Avoid calculating node data more than once when unnecessary
When connecting the same output slot to multiple nodes, each node will get the same data from that output node every time. This means unnecessary calculations, which could cause performance issues.

There are a few ways we could fix this:
- Save the cached data in the node and get it if it's needed
- Save the cached data in a resource shared by all nodes (`GaeaCache` or something).
The first solution could be the simpler one, but for every generation we have to clear the cache, and this wouldn't make that harder to do. The `GaeaCache` resource could instead be created every generation and then freed afterwards, but I'm not sure how well it would play with threading.

Optimization is an expected core feature.

cullumi:
# Caching and Traversal for GaeaData Graph Execution
## Goal
This PR aims to streamline traversal of the Gaea Data graph, as well as implement data caching for eliminating re-calculation of grid data and the like.

Closes #233, and lays some groundwork for #230 

Note:
- It only caches the results of the `get_data` method, not the `get_arg` method.
- The core focus of this PR is `node_resource.gd`, so any method names mentioned are contained in that script.

## How to Test
1. Open project and floor_walker.tres scene.
2. Select logging flags as necessary in Gaea Data resource on Gaea Generator's data field.
3. Click generate for Floor Walker example
4. Note that results are as expected, and logging indicates correct execution, without duplicate runs of any graph nodes.

### Traversal
- Traversal is now handled in the `traverse` method. (it worked exactly the same in every Node Resource)
  - account for required inputs by overriding `_get_required_input_ports` method.
  - `execute` and `preview_texture.gd` now use `traverse` instead of `get_data`.
### Logging
- Logging helpers for confirming that generation is doing what's expected.
  - configure using Logging flags on Gaea Data resource.
  - includes `execute`, `traverse`, `get_data`, and `get_arg` flags.
### Data Retrieval
- Get Data is now simplified, and only handles data calculation, not traversal.
  - `passed_data` is now passed to `get_data` by `traverse`.
  - `get_data` is only called if all required inputs (as returned by `_get_required_input_ports`) are found to be valid.
  - `passed_data` is an Array of Dictionary results corresponding to input port indexes.
### Caching
- GaeaData.cache is where data is stored.
  - this data is stored by Resource Node (`self`), then `output_port` index.
  - it has to be cleared manually by the Generator; this is for better support of executing multiple Output graph nodes.
  - graph node-specific data can be customized by overriding `set_cached_data`, `has_cached_data`, and `get_cached_data`.
### Input Port Validation
- Input Validation for required inputs is handled in `traverse`.
  - has helpers: `has_inputs_connected` and `get_input_resource`.


addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -27,16 +27,84 @@ var node: GaeaGraphNode
...
+	pass
+
+
+# Traversal
+func traverse(output_port:int, area: AABB, generator_data:GaeaData) -> Dictionary:
+	log_traverse(generator_data)
+	
+	# Caching
+	if has_cached_data(generator_data):
+		return get_cached_data(generator_data)
```
BenjaTK: This should be taking into account the output port used, since different output ports return different data.

addons/gaea/graph/nodes/root/data/constant.gd
```diff
@@ -15,7 +16,7 @@ func get_type() -> GaeaGraphNode.SlotTypes:
 func get_icon() -> Texture2D:
 	var for_type: GaeaNodeArgument.Type = args.front().type
 	if for_type == GaeaNodeArgument.Type.FLOAT:
-		return preload("../../assets/types/float.svg")
```
BenjaTK, line 18: These were relative because you can now install addons in a different structure iirc so just doing `res://` doesn't guarantee we'll find these assets. I don't like it but you gotta do what you gotta do 😔

addons/gaea/graph/nodes/root/data/operations/datas_operation.gd
```diff
@@ -1,32 +1,12 @@
 @tool
-extends "res://addons/gaea/graph/nodes/operation.gd"
+extends "res://addons/gaea/graph/nodes/root/data/operation.gd"
 
+func required_slots(): return [0]
```
BenjaTK: I think `required_slots()` should be called `_get_required_slots()`

addons/gaea/graph/nodes/root/map/random/random_scatter.gd
```diff
@@ -1,32 +1,18 @@
...
-			area, generator_data
-		)
-	var material: GaeaMaterial = null
+func get_data(passed_data:Array[Dictionary], _output_port: int, area: AABB, generator_data: GaeaData) -> Dictionary[Vector3i, GaeaMaterial]:
+	log_data(_output_port, generator_data)
+	
+	var grid_data:Dictionary = passed_data[0]
+	var material: GaeaMaterial = passed_data[1].get("value", null)
+	
 	seed(generator_data.generator.seed + salt)
```
BenjaTK, line 12: Should we maybe set the seed in `traverse`? Just so you don't have to keep remembering to use it. (We could even add a `RandomNumberGenerator` variable later on so you don't even have to create that)

BenjaTK (changes requested):
I really like this! I was just about to rewrite how getting the inputs worked, lol. Way too much repeated code before.

addons/gaea/graph/nodes/root/data/constant.gd
Zehir: Why not using the "uid://" paths ?

BenjaTK: I forgot those exist lol

BenjaTK: Do images have them? I thought it was only for resources and scenes.

Zehir: Maybe I could do a PR to update existing path.

I already used it in a previous PR. The tooltip show the real file path and you can ctrl+click to open it
image

Zehir: > Do images have them? I thought it was only for resources and scenes.

Everything have them, the last one that did not was gd script, that why we have .gd.uid file everywhere now.
Script it's since 4.4, image no idea since when

The for image they are stored in the .import file

See https://godotengine.org/article/uid-changes-coming-to-godot-4-4/

cullumi: I see, I see. It was throwing errors with it relative paths (because I moved a few files), but I'll see about using UIDs like Zehir suggested.

addons/gaea/graph/nodes/root/map/random/random_scatter.gd
cullumi: I've actually got another draft PR that was doing stuff with seeds (like making them relative to AABB position, for example). I think we can address this over in #226

addons/gaea/graph/nodes/node_resource.gd
cullumi: Good catch! That's an easy fix; I had left it as just one output since every node I'd seen only had one, but better to support multiple right away rather than discovering this later.

BenjaTK: DecomposeVector2 and DecomposeVector3 are the only ones that have multiple so far I think

cullumi: It should now be handling cached data per output port.

addons/gaea/graph/nodes/root/data/operations/datas_operation.gd
cullumi: I'm realizing I need to clean up a few of these (the one on NodeResource was `required_inputs`), but how about `_get_required_inputs`? Or `_get_required_input_ports` if we wanna be more specific?

BenjaTK: I like `_get_required_input_ports`

addons/gaea/graph/nodes/root/data/constant.gd
cullumi: I'm leaving this to be addressed by #260, though I have completed most of the work of switching these paths over to UIDs.

cullumi:
This one should be all set now. Feel free to give it one last review to sanity check me, or just merge it in. @BenjaTK

BenjaTK (approved):

Issue #202, opened 2025-03-23
Add undo/redo


BenjaTK:

Zehir:
Somebody is already working on it ?
I could work on it, I already made the feature for one of my plugins

PR #253, opened 2025-04-07
Prevent multiple links to a single input

### LINKED ISSUE #221: [2.0] Prevent Multiple Links to a Single Input
# Prevent Multiple Links to a Single Input

## Description
Currently, when an input node has multiple links, the system selects the first connected node as the value source. This behavior can lead to unintended results and should be prevented.

## Steps to Reproduce
1. Create a node with an input.
2. Connect two different output nodes to the same input.
3. Observe that only one of the connected outputs is used, with no clear indication of which one is prioritized.

## Expected Behavior
- The system should prevent multiple links to a single input.
- If multiple links are attempted, an error or warning should be displayed to the user.

## Additional Context
As discussed, this behavior is not ideal and could cause confusion. Adding a check to prevent multiple connections would improve clarity and usability.

Zehir:
Implement checks to ensure that a single input cannot be linked to multiple outputs.

You can hold the shift key to replace the existing connection.

Close #221

### DIFF
--- a/addons/gaea/editor/graph_edit.gd
+++ b/addons/gaea/editor/graph_edit.gd
@@ -32,6 +32,10 @@ func delete_nodes(nodes: Array[StringName]) -> void:
 
 
 func _on_connection_request(from_node: StringName, from_port: int, to_node: StringName, to_port: int) -> void:
+	for connection: Dictionary in get_connection_list():
+		if connection.to_node == to_node and connection.to_port == to_port:
+			disconnect_node(connection.from_node, connection.from_port, connection.to_node, connection.to_port)
+
 	connect_node(from_node, from_port, to_node, to_port)
 	request_connection_update.emit()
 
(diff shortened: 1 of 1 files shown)


Zehir:
I would like to show the user that he can hold the shift key but I don't really know how. Maybe a tooltip showing why he can't connect to a node. Or some kind of status bar on the bottom ?

BenjaTK:
What if instead connecting the node just overrode the connection without holding the shift key? I'd argue that makes more sense and doesn't need a tooltip. It would tell the user intuitively that you can't connect more than one node at the same time.

Zehir:
Yes I did think about that too but I tell myself that if it's was a long wire you may don't know what was connected before.
Maybe the undo/redo from #202 could help ?

BenjaTK:
Yeah I think that will definitely help. I may try to implement undo/redo later today.

BenjaTK:
So we've decided on making it override the connection by default, right?

cullumi:
> So we've decided on making it override the connection by default, right?

That makes the most sense to me! 😄 

> Maybe the undo/redo from 202 could help ?

I think the Undo/Redo feature will make any potentially surprising or confusing disconnects a non-issue.

Zehir:
Updated, we can now always replace the existing link

Zehir:
I had to include the changes made by this PR in #259

PR #257, opened 2025-04-07
Fix generator data assignment to prevent null reference errors

Zehir:
Ensure proper handling of null values in generator data assignments to avoid null reference errors during operations.

This allow to create the `GaeaData` from the inspector and properly reload the Panel.

Also center the view to the output node on the first init of the `GaeaData`

### DIFF
--- a/addons/gaea/editor/panel.gd
+++ b/addons/gaea/editor/panel.gd
@@ -1,7 +1,7 @@
 @tool
 extends Control
 
-var _selected_generator: GaeaGenerator = null : get = get_selected_generator
+var _selected_generator: GaeaGenerator = null: get = get_selected_generator
 var _output_node: GraphNode
 
 const _LinkPopup = preload("uid://btt4eqjkp5pyf")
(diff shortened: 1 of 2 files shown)


addons/gaea/editor/panel.gd
```diff
@@ -34,26 +34,27 @@ func _ready() -> void:
 func populate(node: GaeaGenerator) -> void:
 	_remove_children()
 	_output_node = null
+	if _selected_generator != null and _selected_generator.data_changed.is_connected(_on_data_changed):
```
BenjaTK: Just for cleanliness I like to use `is_instance_valid` instead of `!= null`

addons/gaea/nodes/generator.gd
```diff
@@ -14,7 +14,8 @@ signal area_erased(area: AABB)
 @export var data: GaeaData :
 	set(value):
 		data = value
-		data.generator = self
+		if data:
```
BenjaTK: Same here, `if is_instance_valid(data)`

addons/gaea/editor/panel.gd
Zehir: Updated

addons/gaea/nodes/generator.gd
Zehir: Done too

addons/gaea/editor/panel.gd
```diff
@@ -178,7 +176,7 @@ func update_connections() -> void:
 
 
 func _save_data() -> void:
-	if not is_instance_valid(_selected_generator):
+	if not is_instance_valid(_selected_generator) or _selected_generator.data == null:
```
BenjaTK: You can use `is_instance_valid` here too.

Zehir: Dammit I missed one, I did ctrl+F for != null 😅

PR #259, opened 2025-04-07
Implement reroute node functionality

### LINKED ISSUE #221: [2.0] Prevent Multiple Links to a Single Input

### LINKED ISSUE #247: Add Reroute node

Zehir:
## ✨ Introduce Reroute Node for Enhanced Graph Editing

Closes #247  
Closes #221  
Suppress #253  

This PR introduces a new **Reroute Node** to the graph editor.  
Inspired by the Shader Editor's reroute node, it offers a cleaner and more intuitive editing experience.

### ✅ Features
- Adds a **Reroute Node**.
- Under the hood, it's simply a node that does nothing except pass data through — with one input and one output, of the same type as the link it was inserted on.
- Supports all types of links.
- Mimics the behavior of the Shader Editor's reroute system.
- Prevents users from creating circular connections (loop prevention).
- Automatically replaces existing input connections when inserting a reroute node.
- Automatically reconnects wires if the reroute node is deleted.
- 🖱️ **Node creation UX improvement**: newly created nodes are now placed at the position of the initial click (when right-clicking), rather than where the mouse happens to be when clicking "Add" in the popup.
- 🧰 **New helper method**: added `_local_to_grid(position: Vector2)` in `panel.gd` to convert a local position to a snapped grid position, for consistent node placement.
- 🧰 **New virtual method**: added `_use_caching(output_port:int, generator_data:GaeaData)` in `resource_node.gd` allow a node to not be cached.
- Display mouse position on the bottom left corner
- Added `GaeaNodeResource.log_error(message:String, generator_data:GaeaData, node_id: int = -1)` method to log errors in the console.

#### Reroute node
demo 
#### Multiple reroute types
image
#### Connection limited to avoid loops
demo
#### Missing links on reroute node
image


addons/gaea/editor/link_popup.gd
```diff
@@ -3,12 +3,15 @@ extends PopupMenu
...
 enum Action {
 	DISCONNECT,
+	INSERT_NEW_REROUTE
 }
 
 @export var graph_edit: GraphEdit
 
 var current_connection: Dictionary
 
+signal create_new_reroute(connection: Dictionary)
```
BenjaTK: Signals should be in past tense (`new_reroute_created` or `new_reroute_requested`)

addons/gaea/editor/panel.gd
```diff
@@ -380,3 +385,38 @@ func _on_window_close_requested(original_parent: Control, window: Window) -> voi
 	window.queue_free()
 	_window_popout_button.show()
 	_window_popout_separator.show()
+
+
+func _on_create_new_reroute(connection: Dictionary) -> void:
```
BenjaTK: I think `new_reroute_requested` would make more sense for the signal (this would be `_on_new_reroute_requested`)

Zehir:
@BenjaTK This one is ready for review. I fell into a rabbit hole multiple times, but I think it's a good improvement and will help users make fewer wiring errors.

cullumi:
@Zehir your thinking is correct; the reroute node by default will cache data again.

One option is to simply override the "has_cached_data" method to always return false.

The slightly harder approach is overriding the traverse method for the reroute node.

Zehir:
Yes both could works but i dont see an other node who not use cache but that may change in the future

I first replace the traverse method but that a lot of code to copy

Zehir:
@cullumi And about the go to nowhere issue ?

Zehir:
I have updated the floor walker demo to add reroute nodes;
Before :
image

After:
image

Sadly the git diff is not really good with this resource

Zehir:
I have updated the code to resolve the last issue, now when there is a missing like we get a red circle around the node;

image

And on generation he will use the default value and give a error in the Output console;
  ERROR: res://scenes/tmp_reroute.tscn::Resource_5wqc4:(-460.0, 120.0) in node 'Reroute (Number)' - Could not get data from previous node, using default value instead.

Also now display the mouse position in grid position on the GraphEdit;
image

addons/gaea/graph/nodes/reroute_node.gd
```diff
@@ -0,0 +1,166 @@
...
+func _set_icon_opacity(value: float):
+	if is_instance_valid(tween) and tween.is_running():
+		tween.kill()
+	tween = create_tween()
+	tween.tween_property(self, "icon_opacity", value, 0.3)
+#endregion
+
+
+func _validate_connections():
+	print("_validate_connections")
```
BenjaTK: Remove this print

BenjaTK (changes requested):

addons/gaea/graph/nodes/reroute_node.gd
Zehir: I suppose I forgot to remove it thanks

addons/gaea/editor/link_popup.gd
Zehir: Updated

addons/gaea/editor/panel.gd
Zehir: Updated

PR #268, opened 2025-04-09
Add MaterialGradient and GradientMapper

BenjaTK:
Adds 2 things:
- The GaeaMaterialGradient resource. It holds a list of materials and offsets and can be sampled at a specific point from 0.0-1.0 to get the material corresponding to that point. Useful for mapping noise to terrain types!
Peek 2025-04-09 16-36
- The GradientMapper
Takes a gradient and maps any values from `Data` to the corresponding material.
What previously looked something like this:
image
now looks like this:
image


addons/gaea/assets/slots/diamond.svg.import
```diff
@@ -33,5 +34,5 @@ process/hdr_clamp_exposure=false
 process/size_limit=0
 detect_3d/compress_to=1
 svg/scale=1.0
-editor/scale_with_editor_scale=false
-editor/convert_colors_with_editor_theme=false
+editor/scale_with_editor_scale=true
+editor/convert_colors_with_editor_theme=true
```
Zehir: You need this to be disabled from the Import tab -> Editor or the image won't have the correct size
image

scenes/walker_demo/floor_walker.tres
```diff
@@ -17,7 +17,7 @@
 [ext_resource type="Script" uid="uid://dvbwtw53ut2hc" path="res://addons/gaea/graph/nodes/root/data/complement.gd" id="14_oc7x7"]
 [ext_resource type="Script" uid="uid://bmxbe51xh0t7d" path="res://addons/gaea/graph/nodes/root/data/union.gd" id="17_drb3t"]
 [ext_resource type="Script" uid="uid://br8gcsyc04ksj" path="res://addons/gaea/graph/nodes/root/map/placing/rules_placer.gd" id="17_x5s41"]
-[ext_resource type="Script" uid="uid://b38syakgm25ya" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="18_vyp3c"]
+[ext_resource type="Script" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="18_vyp3c"]
```
Zehir, line 21: Why this uid keep getting removed ?

addons/gaea/graph/nodes/node.gd
```diff
@@ -7,7 +7,7 @@ const PreviewTexture = preload("res://addons/gaea/graph/nodes/preview_texture.gd
 const PREVIEW_TYPES := [SlotTypes.MAP_DATA, SlotTypes.VALUE_DATA]
 
 enum SlotTypes {
-	VALUE_DATA, MAP_DATA, TILE_INFO, VECTOR2, NUMBER, RANGE, BOOL, VECTOR3, NULL = -1
+	VALUE_DATA, MAP_DATA, TILE_INFO, VECTOR2, NUMBER, RANGE, BOOL, VECTOR3, GRADIENT, NULL = -1
```
Zehir: I think we should have one item per line to avoid merge conflict in the future

addons/gaea/graph/nodes/root/data/filters/random_filter.tres
```diff
@@ -2,7 +2,7 @@
 
 [ext_resource type="Script" uid="uid://bh3sajxnmp14w" path="res://addons/gaea/graph/nodes/node_argument.gd" id="1_5n5am"]
 [ext_resource type="Script" uid="uid://ote622wl6axc" path="res://addons/gaea/graph/nodes/node_slot.gd" id="2_trkfv"]
-[ext_resource type="Script" uid="uid://b38syakgm25ya" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="3_5n5am"]
+[ext_resource type="Script" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="3_5n5am"]
```
Zehir, line 5: UID fanished

addons/gaea/graph/nodes/root/map/filters/random_filter.tres
```diff
@@ -2,7 +2,7 @@
 
 [ext_resource type="Script" uid="uid://bh3sajxnmp14w" path="res://addons/gaea/graph/nodes/node_argument.gd" id="1_vr2bg"]
 [ext_resource type="Script" uid="uid://ote622wl6axc" path="res://addons/gaea/graph/nodes/node_slot.gd" id="2_fraxp"]
-[ext_resource type="Script" uid="uid://b38syakgm25ya" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="3_5ax8o"]
+[ext_resource type="Script" path="res://addons/gaea/graph/nodes/root/data/filters/random_filter.gd" id="3_5ax8o"]
```
Zehir, line 5: UID vanished

BenjaTK: Nothing I can do about this, idk why Godot does it.

Zehir: Can you revert it with the git editor ? Or it's will collide with #259
We will see after the 2 merges if it's still happening

BenjaTK:
UID problem(?) couldn't be solved since Godot just does it automatically

Zehir:
With no defined resource in the GaeaData I get this error ;
  ERROR: res://addons/gaea/graph/nodes/root/map/mappers/gradient_mapper.gd:16 - Invalid call. Nonexistent function 'sample' in base 'Nil'.

With the same case for the MaterialVariable and the Mapper I get nothing but no error

Zehir:
For me the order of the offset are reversed;
In the array they are from bottom to top and on the visual colors checkerboard they are from left to right.
For me it's should be in the array from 0.0 to 1.0

Zehir:
When you are re ordering the offsets the materials order is modified too but when changing the order of the material the offset don't.
Do we really want auto reorder of the material when updating the offsets ?

Zehir (commented):
With no defined resource in the GaeaData I get this error ;
  ERROR: res://addons/gaea/graph/nodes/root/map/mappers/gradient_mapper.gd:16 - Invalid call. Nonexistent function 'sample' in base 'Nil'.

With the same case for the MaterialVariable and the Mapper I get nothing but no error


For me the order of the offset are reversed;
In the array they are from bottom to top and on the visual colors checkerboard they are from left to right.
For me it's should be in the array from 0.0 to 1.0


When you are re ordering the offsets the materials order is modified too but when changing the order of the material the offset don't.
Do we really want auto reorder of the material when updating the offsets ?

BenjaTK:
All those should be fixed

PR #272, opened 2025-04-10
Add output preview panel

### LINKED ISSUE #264: Output preview
It's would be nice to have a side panel like the ShaderGraph to display the layers on the right side.

BenjaTK:
Peek 2025-04-10 12-47

Adds a preview of the whole output to the dock, including a way to preview specific layers, change the axis mode and change the tertiary axis position.

Closes #264


cullumi:
> I wonder if it would be too much of a performance drain to make this a viewport instead and use a GridMap for 3D generations? It sounds like it would be. @cullumi @Zehir 

I could totally see using a viewport for that. I don't think it'd be too much of a performance drain. Depends exactly what you're doing inside that viewport, I guess.

BenjaTK:
Stale

Zehir:
This was a nice feature, do you think we can salvage it @BenjaTK  ?

PR #275, opened 2025-04-10
Rename nodes and ensure unique access in panel

Zehir:
Rename nodes in the panel for clarity and implement unique names for better access and management.


Zehir:
I renamed the Tree node to match the script path

PR #287, opened 2025-04-12
Add vector2 and range casting functionality

Zehir:
Introduce methods for casting between VECTOR2 and RANGE types, enhancing type compatibility in the graph node system.

More cast to go but I would like your feed back on that one and will do other PR for other casts;

image

### DIFF
--- a/addons/gaea/graph/nodes/node_resource.gd
+++ b/addons/gaea/graph/nodes/node_resource.gd
@@ -150,14 +150,7 @@ func get_arg(name: String, generator_data: GaeaData) -> Variant:
 				var connected_type := connected_node.get_output_port_type(connected_port)
 				if arg_slot_type == connected_type:
 					return connected_data.get("value")
-
-				var method_name = &"cast_from_%s_to_%s" % [
-					GaeaGraphNode.SlotTypes.find_key(connected_type).to_snake_case(),
-					GaeaGraphNode.SlotTypes.find_key(arg_slot_type).to_snake_case(),
-				]
-				if has_method(method_name):
-					return call(method_name, connected_data.get("value"))
-				log_error("Could not get data from previous node, missing cast method : %s" % method_name, generator_data, connected_idx)
+				return GaeaNodeResource.cast_value(connected_type, arg_slot_type, connected_data.get("value"))
 			else:
 				log_error("Could not get data from previous node, using default value instead.", generator_data, connected_idx)
 	return data.get(name)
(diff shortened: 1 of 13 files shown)


Zehir:
This handle the case where it's an arg that need to be casted but I am not sure about the input cast.

@cullumi can you take a look please ?

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -126,23 +126,38 @@ func get_arg(name: String, generator_data: GaeaData) -> Variant:
...
 				AABB(),
 				generator_data
 			)
 			if connected_data.has("value"):
-				return connected_data.get("value")
+				var connected_type := connected_node.get_output_port_type(connected_port)
+				if arg_slot_type == connected_type:
+					return connected_data.get("value")
+
+				var method_name = &"cast_from_%s_to_%s" % [
```
BenjaTK: I think it makes more sense to have a `cast_value(from_type, to_type, value)` method instead. I don't like using method names since they can be finicky and you don't get errors until you run it.

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -279,3 +334,19 @@ func _is_point_outside_area(area: AABB, point: Vector3) -> bool:
 	return (point.x < area.position.x or point.y < area.position.y or point.z < area.position.z or
 			point.x > area.end.x or point.y > area.end.y or point.z > area.end.z)
 #endregion
+
+
+#region Data casting methods
+func cast_from_range_to_vector_2(value: Dictionary) -> Dictionary:
```
BenjaTK: The Vector2 type is passed like an actual Vector2.
```
{
    "value": Vector2(x, y)
}
```

BenjaTK (changes requested):

addons/gaea/graph/nodes/node_resource.gd
Zehir: You prefer a big switch bloc with the code in it or just a big switch bloc that call all methods ?

addons/gaea/graph/nodes/node_resource.gd
BenjaTK: This method should instead return a Vector2

Zehir: Yes right, first I was thinkig to pass the {value:"thing"} Dictionary

addons/gaea/graph/nodes/node_resource.gd
BenjaTK: I don't really mind

Zehir: Added in latest commit

addons/gaea/graph/nodes/node_resource.gd
Zehir: Added in latest commit

BenjaTK (approved):
LGTM

Zehir:
I added the casting for input even if for now I found no nodes that have a vector or range input that is not an arg but could be usefull in the future

BenjaTK:
None will have. They'll always be args, only ones that are only inputs are stuff like Data, resources and Map that don't have an interface.

BenjaTK:
But it's fine to have

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -44,25 +44,31 @@ func traverse(output_port:int, area: AABB, generator_data:GaeaData) -> Dictionar
 
 	# Validation
 	if not has_inputs_connected(_get_required_input_ports(), generator_data):
+		log_error("The node don't have all required inputs ports connected.", generator_data, generator_data.resources.find(self))
```
BenjaTK: I'm not sure if this should produce an error, I'd say lets leave it for another issue and PR

BenjaTK (changes requested):

addons/gaea/graph/nodes/node_resource.gd
Zehir: Ok I removed the log error

PR #290, opened 2025-04-12
Add online docs button

Zehir:
This button is already existing in the Script editor and will be in 4.5 on the Shader Graph editor so I made a Gaea version.

image


BenjaTK:
I think it should be moved to the right/left of the Window Popout button and not hide the separator when popping out the panel

Zehir:
I don't understand what you mean, I just copy the 4.5 version of the shader graph
image

PR #292, opened 2025-04-13
Add more value casting

BenjaTK:
image

- `num —> Vector2/3 = Vector2(value, value)/Vector3(value, value, value)`
- `Vector2 —> Vector3 = Vector3(x, y, 0.0)`
- `Vector3 —> Vector2 = Vector2(x, y)`
- `bool <—> num = 1.0 <—> true, 0.0 <—> false`

### DIFF
--- a/addons/gaea/graph/nodes/node_resource.gd
+++ b/addons/gaea/graph/nodes/node_resource.gd
@@ -338,14 +338,14 @@ func _is_point_outside_area(area: AABB, point: Vector3) -> bool:
 #region Data casting methods
 static func cast_value(from_type: GaeaGraphNode.SlotTypes, to_type: GaeaGraphNode.SlotTypes, value: Variant) -> Variant:
 	match [from_type, to_type]:
+		#region Range -> Any
 		[GaeaGraphNode.SlotTypes.RANGE, GaeaGraphNode.SlotTypes.VECTOR2]:
 			return Vector2(
 				value.get("min"), value.get("max")
 			)
-		[GaeaGraphNode.SlotTypes.VECTOR2, GaeaGraphNode.SlotTypes.RANGE]:
-			return {
-				"min": value.x, "max": value.y
-			}
+		#endregion
+
+		#region Number -> Any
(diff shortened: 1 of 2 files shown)


addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -340,14 +340,33 @@ static func cast_value(from_type: GaeaGraphNode.SlotTypes, to_type: GaeaGraphNod
...
+				value.x, value.y, 0.0
+			)
+		[GaeaGraphNode.SlotTypes.VECTOR3, GaeaGraphNode.SlotTypes.VECTOR2]:
+			return Vector2(
+				value.x, value.y
+			)
+		[GaeaGraphNode.SlotTypes.BOOL, GaeaGraphNode.SlotTypes.NUMBER]:
+			return (1.0 if value else 0.0)
+		[GaeaGraphNode.SlotTypes.NUMBER, GaeaGraphNode.SlotTypes.BOOL]:
+			return is_equal_approx(value, 1.0)
```
Zehir: Usually, to check if it's a truthy value, we use value > 0

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -365,7 +365,7 @@ static func cast_value(from_type: GaeaGraphNode.SlotTypes, to_type: GaeaGraphNod
 		[GaeaGraphNode.SlotTypes.BOOL, GaeaGraphNode.SlotTypes.NUMBER]:
 			return (1.0 if value else 0.0)
 		[GaeaGraphNode.SlotTypes.NUMBER, GaeaGraphNode.SlotTypes.BOOL]:
-			return is_equal_approx(value, 1.0)
+			return value > 0.0
```
Zehir: Now that I think again about it and because float are float maybe it's better to do;
```gdscript
return not is_zero_approx(value)
```

BenjaTK: I'll just do `bool(value)`

Zehir (approved):
Yes that works too I think

LGTM

PR #293, opened 2025-04-13
Hide arguments when input is connected

### LINKED ISSUE #242: Hide arguments inside nodes when input is connected
Image

The value inputs here should be hidden when the slot is connected to something.

BenjaTK:
image

Closes #242

### DIFF
--- a/addons/gaea/editor/graph_edit.gd
+++ b/addons/gaea/editor/graph_edit.gd
@@ -60,7 +60,6 @@ func _on_connection_request(from_node: StringName, from_port: int, to_node: Stri
 					connection.to_port
 				)
 
-	target_node.hide_param_value(to_port)
 	connect_node(from_node, from_port, to_node, to_port)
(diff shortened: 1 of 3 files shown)


BenjaTK:
I'm having a problem I can't solve and I need some help with.
Peek 2025-04-13 14-05
As you can see the wires only update when I zoom out/in.

BenjaTK:
image
One idea I had was to just set the minimum size on the parameters on `_ready` which looks fine in all except the ones that have more than one line

BenjaTK:
Managed to fix it, I just hide the argument before actually connecting the nodes so the GraphEdit can properly detect where the slot is.
Peek 2025-04-13 14-28

BenjaTK:
Peek 2025-04-13 14-50

The output slot is still misaligned

Zehir (approved):
LGTM

PR #294, opened 2025-04-13
Prioritize updates in the same branch, warn when updating from 1.X to 2.0

Zehir:
Added a version check to prioritize 1.X updates, and if none are available, show the 2.X update with a warning.

### DIFF
--- a/addons/gaea/editor/download_update_panel.tscn
+++ b/addons/gaea/editor/download_update_panel.tscn
@@ -1,7 +1,7 @@
 [gd_scene load_steps=3 format=3 uid="uid://6pmddcctde0v"]
 
 [ext_resource type="Texture2D" uid="uid://b4nwpbrra72fj" path="res://addons/gaea/editor/logo.png" id="1_05fnf"]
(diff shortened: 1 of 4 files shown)


addons/gaea/editor/download_update_panel.tscn
```diff
@@ -1,7 +1,7 @@
 [gd_scene load_steps=3 format=3 uid="uid://6pmddcctde0v"]
 
 [ext_resource type="Texture2D" uid="uid://b4nwpbrra72fj" path="res://addons/gaea/editor/logo.png" id="1_05fnf"]
-[ext_resource type="Script" path="res://addons/gaea/editor/download_update_panel.gd" id="1_moikk"]
+[ext_resource type="Script" uid="uid://wisjqo0n7ge" path="res://addons/gaea/editor/download_update_panel.gd" id="1_moikk"]
```
BenjaTK: 1.X should work in 4.3 but this added an UID

Zehir: yes I missed it but I don't think there is any problem with it but I removed it anyway

PR #295, opened 2025-04-13
Fix reroute node

Zehir:
I am still using the left and right of an input slot.
It's a big rewrite to use an input slot and an output slot so I stick with this quick fix

### DIFF
--- a/addons/gaea/graph/nodes/reroute_node.gd
+++ b/addons/gaea/graph/nodes/reroute_node.gd
@@ -25,6 +25,8 @@ var has_no_input: bool = false:
 #region init
 func initialize() -> void:
 	super()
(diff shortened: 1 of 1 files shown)


Zehir:
Also when droping a link to nothing, instead of openning the add node popup we could add a reroute node instead.
What do you think ?

BenjaTK:
I think opening the add node popup makes more sense. We could consider making it connect to the new node if valid, though. Like the shader editor

BenjaTK:
I think the shader editor filters the create node tree when doing this to only valid nodes

Zehir:
> I think the shader editor filters the create node tree when doing this to only valid nodes

yes he does

PR #298, opened 2025-04-13
Set `resizable` to false on nodes

BenjaTK:


Zehir:
Can you remove the resizable = false in the reroute_node.tscn since it's now set in the base scene ?

BenjaTK:
Done

PR #309, opened 2025-04-16
Add shortcut to cycle between nodes in the 'Create Node' tree

BenjaTK:
Peek 2025-04-16 09-22

`Ctrl + Tab` cycles between search results (and if there's no search, just the whole tree), but only between selectable items. `Ctrl + Shift + Tab` goes the opposite direction. You can then press `Enter` to create the selected node.

This PR also makes categories in the tree unselectable.


Zehir (approved):

cullumi (approved):
Works great! I didn't find any ways to break it, personally.

PR #310, opened 2025-04-16
Add enum parameter and float operation functionality

Zehir:
Introduce an enum parameter for dynamic input handling and add a new float operation, enhancing the functionality of the graph node parameters.

The list is generated from a enum on the resource that have the same name capitalized.
The enum parameter should not have any input because that won't make sense

In long term the node will replace each of the nodes like the Add but we will first need to have a way to add multiple entry in the add node list from a single resource file.

Also this work only for vectors but I think we could add an other arg before the operation to support vectors an other stuff like that.

I think this PR is big enought to be a first step

demo


BenjaTK:
Is `get_default_value` defined for enums?

Zehir:
The _get_default_value from the variable_node ?
No because we can't have enum in variable nodes

BenjaTK:
It's what I added in the save system PR but just realized it hasn't been merged yet

Zehir:
I think we should merge both and update what is missing after

BenjaTK:
image
When it first spawns it looks like this

BenjaTK:
Also the enum parameter doesn't go all the way

BenjaTK:
I personally think instead of being X and Y, the parameters should be A and B respectively. Not sure what everyone else thinks about this but it matches the Shader editor language

BenjaTK:
image
the parameter name in the variable nodes are also showing up now when they shouldn't.

Zehir:
> image When it first spawns it looks like this

Dammit it was fixed

Zehir:
> Also the enum parameter doesn't go all the way

Wtf it was. I maybe it'd because the base was not the same, I will rebase the pr to check

Zehir:
> I personally think instead of being X and Y, the parameters should be A and B respectively. Not sure what everyone else thinks about this but it matches the Shader editor language

I used the name from the documentation, most of them are X and Y and some are A or B

Zehir:
I have removed the `_on_static_arg_value_changed` callback in the resource that I added for a more generic approach `get_scene_script` to just replace the script when you don't need a custom scene and moved the related code from `float_op` to `float_op_node`.
And hide labels on the variables

Zehir:
> Also the enum parameter doesn't go all the way

Can you try again with the latest commit ? I have no issue with it

Zehir:
Also if we want this node to replace the existing add node we will need a system to add more than one entry in the create node panel. Do you wan't me to add it to this PR or do an other PR later ?

And an other idea with the new system planned in #311 maybe we can have a smaller list but with more outputs, like you have the "min/max" option and 2 outputs min and max, "rounding" option for round, ceil, floor, and so on.

BenjaTK:
Im not sure what you mean by that second paragraph. And the first point, I think that should be its own PR.

Zehir:
The idea was to have more generic options that have multiple outputs instead of many options with only one output each time

BenjaTK:
I think it's fine to have only one output.

Zehir:
Ok then it's ready for merge

cullumi:
@Zehir, I'm noticing odd behavior when connecting to the second slot then changing to an operation that only accepts one. The slot connection seems to remain (and cannot be disconnected manually), and the following error is thrown when generating:

`ERROR: scene/gui/graph_node.cpp:725 - Index p_port_idx = 1 is out of bounds (left_port_cache.size() = 1).`

This is what it looks like in the graph:
image

Zehir:
> @Zehir, I'm noticing odd behavior when connecting to the second slot then changing to an operation that only accepts one. The slot connection seems to remain (and cannot be disconnected manually), and the following error is thrown when generating:
> 
> `ERROR: scene/gui/graph_node.cpp:725 - Index p_port_idx = 1 is out of bounds (left_port_cache.size() = 1).`
> 
> This is what it looks like in the graph: image

That wierd, but I will redo the PR for #311 anyway because I will replace the base input and output workflow

Zehir:
This PR is completly broken now, I will just close it and will salvage some part of code to make a new PR

PR #312, opened 2025-04-17
Prevent parameter code running in edited scene and fix output node display

Zehir:
Ensure that parameter code does not execute when part of an edited scene. This was causing editor icons to be duplicated in the resources and various other side effects.
Additionally, address issues with the output node display.

This PR need #310 first to work


BenjaTK:
How is this looking after #319?

Zehir:
Idk but probably not good, I will rebase it after

Zehir:
Some of the code changed are in the #319, I put the remaining in #320 and #321

PR #314, opened 2025-04-17
Properly set all previously `null` default values

BenjaTK:


Zehir:
This will collide with #310, I added the _property_get_revert but I keep `get_default_value` but I think now the `get_default_value` should just fetch the value from _property_get_revert

```gdscript
func get_default_value() -> Variant:
	return _property_get_revert("default_value")
```

BenjaTK:
What is `_property_get_revert`?

BenjaTK:
Have you checked that it works with the save system? That property is a Variant, so how can you be sure it's giving the proper default value?

Zehir:
> What is `_property_get_revert`?

It's to set custom values for the "revert" button aka default values;
image

There is 2 methods to make it work;
https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-property-can-revert
https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-property-get-revert

Zehir:
> Have you checked that it works with the save system? That property is a Variant, so how can you be sure it's giving the proper default value?

I did not setup your strict typing but I got the sames changes in the resources, the null values got replaced but I did not include thoses changes in the PR

BenjaTK:
Cool, then I'll close this PR.

Zehir:
I will add your strict types in my PR

PR #319, opened 2025-04-18
Refactor node input and outputs

Zehir:
## Summary

This update reorganizes the node slot system and introduces a unified type system across the plugin.

### Changes

- Previously, nodes used three arrays for slots:
  - `@export var input_slots: Array[GaeaNodeSlot]`
  - `@export var args: Array[GaeaNodeArgument]`
  - `@export var output_slots: Array[GaeaNodeSlot]`
- Now, nodes use two arrays:
  - `@export var params: Array[GaeaNodeSlotParam]`
  - `@export var outputs: Array[GaeaNodeSlotOutput]`
- All existing nodes have been manually updated to use the new slot format.
- Fixed several issues found during the migration, such as `substract_datas` incorrectly using the sum operation.
- Removed `GaeaGraphNode.SlotTypes` and `GaeaNodeArgument.Type` in favor of a new class: `GaeaValue`.
  - `GaeaValue` provides a unified type system shared across the plugin.
  - The class handles icons, colors, types, casting, default values, and more.
- Added a migration function to update reroute nodes, as their stored type was previously invalid.
- Number type does not exist anymore, there is now Float and Int but they have the same color and icon, the user won't see it and the casting method will handle the value conversion.

### To be addressed in a future PR

- Add int vectors params for walker node
- Add a way to show or hide slots.
- Add a way to display param and output on a single line when they have the same name (e.g., for constants or external variables).
- Some casting logic is still missing.


Zehir:
I have updated the preview workflow to clear the cache when editing a node, and to regenerate the edited node along with its descendants. There's also a side effect where the renderer picks up the changes—but I think the bug turned into a feature.
demo

BenjaTK:
> I have updated the preview workflow to clear the cache when editing a node, and to regenerate the edited node along with its descendants. There's also a side effect where the renderer picks up the changes—but I think the bug turned into a feature.
> demo
> 

I really don't like this. Especially with bigger generations this would cause a lot of lag.

BenjaTK:
Also the preview shouldn't generate the whole area. It's restricted only to a smaller size to avoid lag. I think the preview should be reverted to how it worked before.

Zehir:
@BenjaTK we can make it a setting if you want

BenjaTK:
I think this PR should be kept to all commits until https://github.com/gaea-godot/gaea/commit/b1e4437c51cf49568b9f550bcc25216c22b1fc47

BenjaTK (changes requested):
Please revert any commits after https://github.com/gaea-godot/gaea/commit/b1e4437c51cf49568b9f550bcc25216c22b1fc47, otherwise the whole refactor looks really good :fire: . Will take a look at the actual code later.

Zehir:
I revert but the preview was broken at that point in time, I will add a little commit to fix it

Zehir:
Ok now it's ready for review

Zehir:
Just did last minute cleanup, I think I forgot nothing this time

addons/gaea/graph/nodes/node_slot_param.gd
```diff
@@ -0,0 +1,124 @@
...
+	match property:
+		&"default_value":
+			return GaeaValue.get_default_value(type)
+	return super(property)
+
+
+func _validate_property(property: Dictionary) -> void:
+	if property.name == "default_value":
+		match type:
+			GaeaValue.Type.FLOAT:
```
Zehir: Do you think that the property hint code here should me moved in the GaeaValue class ?

addons/gaea/graph/components/graph_node_output.gd
```diff
@@ -0,0 +1,39 @@
+@tool
+class_name GaeaGraphNodeOutput
+extends MarginContainer
+
+const PREVIEW_TYPES := [GaeaValue.Type.MAP, GaeaValue.Type.DATA]
```
BenjaTK: This array isn't used anywhere anymore

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -129,103 +113,74 @@ func get_input_resource(slot:int, generator_data:GaeaData) -> GaeaNodeResource:
...
+	var param := find_param_by_name(name)
+	if not is_instance_valid(param):
+		return null
 
-	var arg_resource: GaeaNodeArgument
-	for arg in args:
-		if arg.name == name:
-			arg_resource = arg
-			break
+	var connexion := get_param_connection(param)
```
BenjaTK: connection :stuck_out_tongue_closed_eyes:

addons/gaea/graph/nodes/node_slot_param.gd
BenjaTK: > Do you think that the property hint code here should me moved in the GaeaValue class ?

I think it could, yeah

addons/gaea/graph/nodes/preview_texture.gd
```diff
@@ -51,34 +53,34 @@ func _ready() -> void:
...
 
 
-func toggle(for_idx: int, for_type: GaeaGraphNode.SlotTypes) -> void:
+func toggle(for_output: GaeaNodeSlotOutput) -> void:
 	if not get_parent().visible:
 		get_parent().show()
-		output_idx = for_idx
-		slider_container.visible = for_type == GaeaGraphNode.SlotTypes.DATA
-		type = for_type
+		slider_container.visible = true
```
BenjaTK: The slider should be visible only if the type is data, since it doesn't make sense in the map preview

BenjaTK (dismissed):

addons/gaea/graph/components/graph_node_output.gd
Zehir: That wierd I remember removing it

addons/gaea/graph/nodes/node_resource.gd
Zehir: Little part of my country 😀

BenjaTK:
LGTM

addons/gaea/graph/components/inputs/graph_node_parameter.gd
```diff
@@ -24,23 +22,26 @@ func initialize(_graph_node: GaeaGraphNode, _slot_idx: int) -> void:
...
-		GaeaEditorSettings.get_configured_icon_for_slot_type(input_type), GaeaEditorSettings.get_configured_icon_for_slot_type(input_type),
-	)
+	
+	if GaeaValue.is_wireable(resource.type):
+		graph_node.set_slot_enabled_left(slot_idx, true)
+		graph_node.set_slot_type_left(slot_idx, resource.type)
+		graph_node.set_slot_color_left(slot_idx, GaeaValue.get_color(resource.type))
+		graph_node.set_slot_custom_icon_left(slot_idx, GaeaValue.get_slot_icon(resource.type))
+	else:
+		# This is required because without it the color of the slots after is OK but not the icon.
```
Zehir, line 42: @BenjaTK when making this PR I had the same issus as in #404 I am on mobile rn and for some reason I can't copy a link to this line of code but you should be able to get a link from this comment.
The idea was to disable the left slot, because doing nothing was ok for the colors but not for the icons. This look verry similar to the other issue. I think this part was removed in the other rework

PR #320, opened 2025-04-19
Cleanup parameter scenes

Zehir:
Some scene have saved the editor icons in their resources. They are not used because they get replaces during loading.

This PR have #319 as base, that why the log is scarry.


BenjaTK:
Could you cherry pick instead the last commit please?

Zehir:
I was thinking the files was edited by #319 but it's not the case so yes I can do that

Zehir:
Ok it's rebased

PR #321, opened 2025-04-19
Fix output node label

Zehir:
Then the layer did not have a label the default name was not dim to gray to show the disabled state


BenjaTK:
Has some conflicts that need to be resolved

Zehir:
No not resolved

PR #326, opened 2025-04-19
Add documentation to most scripts and update naming of certain symbols

BenjaTK:
Closes #317 
Fixes #338
Some of the documentation is added in form of normal comments so said private methods or variables don't clutter the documentation page of the class.  

Online tutorials also added when corresponding.

This PR also updates the naming of certain methods to make them private (`_` prefix), updates the naming of some signals and variables, updates the naming of some classes (e.g. `GaeaGraphNodeParameter` -> `GaeaGraphNodeParameterEditor`) and more. It also adds `class_name` to most scripts, such as all the `GaeaNodeResource`s.

<details><summary>Examples</summary>
<p>

image

image

image

</p>
</details>

PR #329, opened 2025-04-20
Fix runtime generation #306

### LINKED ISSUE #306: Reroute node casting during runtime
**Describe the bug**

For some reason I get this error during only runtime;
  ERROR: Could not get data from previous node, missing cast method from MAP to DATA
  ERROR: Could not get data from previous node, missing cast method from MAP to DATA
  ERROR: Could not get data from previous node, missing cast method from MAP to DATA

It's caused only by the reroute node on green wire only.

I will have to take a look on it but after merge of #305 to avoid merge conflicts

Zehir:
Currently, runtime generation does not work due to the large PR I made earlier.

Fixes #306


Zehir:
Salvaged into #351

PR #330, opened 2025-04-20
Add generation button on demo

Zehir:

This PR add a generation button on the top right corner of the demo scene to generate a new map.
Also change the world size to make it fit the screen


Zehir:
Salvaged into #351

PR #339, opened 2025-04-20
Support for non square tilemaplayers

Zehir:

This PR goal is to update the tilemap renderer to place the cell correctly depending of the used tileset


Zehir:
I just think about this now but maybe we could put the position conversion inside a node instead of the renderer ?

Zehir:
Also the support is not complete, it's just for the renderer, we need an option on the GaeaGraph with the tile layout or on the neighbours, border, ... nodes to move the checkbox and the process correctly

Zehir:
what do you think, we add a setting on the node or directly on the graph that will affect all nodes ?
image
There is also the walker that need more directions ; 
imageimage

I think a node option make more sense or a new node that copy most of the code

Zehir:
@BenjaTK I have a better idea to handle this. I can make a Data and Map position conversion node between 2 layout. The node will have 2 enums to set the source and target layout and you will just need to select the correct conversion to make the data in the expected layout for your tilemap.
This will handle the case where the tilemap is not in the default layout and for isometric tiles.

But there is still the problem of non square tiles, should we add instead a "GaeaData" property with the layout and every node on the graph would update for the new layout ?
Or we make variation nodes for the other tile layout, with the new system it's will not be that hard to handle.

What do you think we should do for this issue ?

BenjaTK:
> @BenjaTK I have a better idea to handle this. I can make a Data and Map position conversion node between 2 layout. The node will have 2 enums to set the source and target layout and you will just need to select the correct conversion to make the data in the expected layout for your tilemap.
> This will handle the case where the tilemap is not in the default layout and for isometric tiles.
> 
> But there is still the problem of non square tiles, should we add instead a "GaeaData" property with the layout and every node on the graph would update for the new layout ?
> Or we make variation nodes for the other tile layout, with the new system it's will not be that hard to handle.
> 
> What do you think we should do for this issue ?

I think I like the idea of the conversion node, but adding the `GaeaGraph` property kinda conflicts with it. It's 2 different properties doing the same. I think I prefer the graph one?

Zehir:
> I think I like the idea of the conversion node, but adding the `GaeaGraph` property kinda conflicts with it. It's 2 different properties doing the same. I think I prefer the graph one?

Yes it's 2 direction for the PR, should be one or the other. I think the "better" way it to have a enum on the nodes that require it like the neighbor. And having a simple node to convert the coordinates of the data that the user will need to put before the output node. Could be data or mapped data (white or green)

BenjaTK:
> > I think I like the idea of the conversion node, but adding the `GaeaGraph` property kinda conflicts with it. It's 2 different properties doing the same. I think I prefer the graph one?
> 
> Yes it's 2 direction for the PR, should be one or the other. I think the "better" way it to have a enum on the nodes that require it like the neighbor. And having a simple node to convert the coordinates of the data that the user will need to put before the output node. Could be data or mapped data (white or green)

The thing about the per-node solution is it's a lot of configuration and more. I think it'd make sense to have a property in `GaeaGraph` and have the different argument editors and nodes adapt to it.

BenjaTK:
The conversion could be done automatically, for example. In the output node.

Zehir:
> The conversion could be done automatically, for example. In the output node.

The problem with that is for isometric tilemap, where is the north ? Top right or top left ? A setting on the gaeageaph ?

Zehir:
Or a node to edit the data keys like rotate, flip, mirror, shift, ...

BenjaTK:
There's a lot of changes here that aren't related to the PR

Zehir:
Yes I need to rewrite the PR, I will wait the rework of the save system you are working on and will redo the PR with the new system we discuss here

solodeveloping:
ExampleIsometric
What about rotating the RulesPlacer, for isometric diamond, the grid exist, it's just difficult to figure out how to use the rules

Zehir:
> What about rotating the RulesPlacer, for isometric diamond, the grid exist, it's just difficult to figure out how to use the rules

It's not just for the isometric that need ajustement of nodes that have a "neighbor" setting. But I don't know, I think for this case I would keep the layout like this and just have a setting at the end before render

Zehir:
Since May, a lot of things have changed inside Gaea. I salvaged what I could from the old PR and force-pushed it here.

Regarding the nodes with checkboxes, like Border2D and RulesPlacer, we have a few options for handling the hexagon layout:

1. Create a hexagon layout variant for each node. (But that could be many nodes)
2. Add a graph-wide setting that updates all nodes. (But you will have many specific code everywhere)
3. Add an enum parameter in each node to select the layout. (Will have 2 version of some methods inside)

What do you think, @BenjaTK and @cullumi?

Personnaly I think 3 is the best idea but maybe you have a 4th one ?

cullumi:
Would it make sense to treat hex vs square data as different "data types"? Like, if you pass hex grid data as a node argument, that node should intuit that it is hexagonal rather than square?

Other nodes that couldn't intuit data in that way (such as generators) would have a dropdown enum.

We could even have a layout conversion node or something. Who knows, maybe processing grid data in one layout then transforming to the other would be useful for whatever reason.

This would also allow for having tilemap layers that have different grid layouts in the same graph (for whatever purpose the user might want that, lol).

Zehir:
For me the coordinate conversion should be done by the renderer, it's currently what this PR does because it's dependant of the tilemap. And we keep one coordinate system for all nodes. Because if you are starting to have multiple coordinate system in the same graph you will need to convert multiples times and it's will be a mess.
So just a enum for the node to know how to handle the neighbors, could be just 4 sides but could be 6 neighbors when there is an offset.

For the node POV you don't really case the shape of the tile or the tileset. It's will be always a grid with cell and coordinates that go up toward bottom right corner.
It's the role of the tilemap renderer to convert that into the used tileset coordinate system.

I also like the idea to be able to do stuff on the key like map rotation, flip, shift, ... but it's not the goal of this PR.

For me the only option is to have a enum with an Horizontal Offset, Vertical Offset or no offset.
Horizontal Offset :
image
image
image

Vertical Offset :
image
image
image

BenjaTK:
I think whatever can be done with less friction and work needed from user the best, but I won't have much of an opinion on this until I see a mock-up.

Zehir:
OK then I will do what I think  is best and we will start from that

Zehir:
I have reduced the scope of this PR, now it's just about the tilemap renderer that now render the cells correctly for all tileset format.

cullumi (approved):
Appears to work properly, decent solution too.

Zehir:
> Appears to work properly, decent solution too.
Thanks !
You will to trust me on the math but it's was used since many months in my other plugin and I got no issues raised by users.

PR #344, opened 2025-04-23
Rework `GaeaNode`s to be configured from their script instead of `.tres` files

### LINKED ISSUE #327: Clarify the naming of parameters, variables and arguments
These 3 terms mean very different things across the codebase and it's confusing.

- Arguments should be the things inside the node. So `GaeaNodeSlotParam` would be renamed into `GaeaNodeSlotArgument`.
- Variables should just be named parameters. In `GaeaData` we have the `parameters` dictionary and I think that's the naming we should use for the variable nodes.

I considered renaming the `parameters` dictionary to `variables` but that would need its own migration feature for saved data so it'd be a mess. We could do it, we just would either have to add that or let it "break".

### LINKED ISSUE #328: Re-organize the folder structure of the addon.
Some things are all over the place, and there's some folders that are starting to have way too many scripts.

I think just some general cleaning, adding more sub-folders and more should do the job.

### LINKED ISSUE #332: 2 Assert need to be removed
Theses 2 asserts need to be removed;

This one is an issue because the name is no longer `border` but `data`.
https://github.com/gaea-godot/gaea/blob/d5c294d6bc928e40a485812b945b26f9f78b4a26/addons/gaea/graph/nodes/root/data/border/border.gd#L11

This one is the only remaining one
https://github.com/gaea-godot/gaea/blob/d5c294d6bc928e40a485812b945b26f9f78b4a26/addons/gaea/graph/nodes/root/data/basic/fill.gd#L7

Can't do PR now because it's will conflicts with #326 (I tried)

### LINKED ISSUE #333: Output label should be rich text label with bbcode
Image

### LINKED ISSUE #341: Customise display of params and outputs
Hello,

Not an issue but I would like your feedback on my idea.

### Context
Here the Constant node should not have a input connection. It's working but it's does not really make sense.
Image

There is also the case where the output result is now just "Data" instead of complement
Image

And sometime you just don't want to display the output label and merge it with the field above;
Image

Also to do complexe nodes like this ;
demo
(From #310)

So the problem, is where to add thoses features ?

1. We could add a boolean value on the GaeaNodeSlotParam to hide the connection point. But it's will be almost never used.
2. We could add a method to resource `_get_param_vibility(param: GaeaNodeSlotParam) -> bool` to be able to hide the complete slot.
3. We could add a method to resource `_get_param_label(param: GaeaNodeSlotParam) -> String` to be able to have a custom label.
4. We could add a method to resource `_get_output_vibility(output: GaeaNodeSlotOutput) -> bool` to be able to hide the complete slot.
5. We could add a method to resource `_get_output_label(output: GaeaNodeSlotOutput) -> String` to be able to have a custom label.
6. We could add the methods 2 to 5 on the node script but you would need to extend the script `GaeaGraphNode` for each node that will need to use custom display. Currently only the Reroute, Output and Variable node does that.
7. Instead of the methods 2 to 5 we could have a new array of "GaeaNodeSlotVisibility" (on the resource) who just be a list of rules to show / hide / set label for Param and Output. You will have a "If" part and a "Then" part on that resource.
8. Add a Script property where we can side load script that will have the methods 2 to 5 but without the need to have to use the _get_scene() method with a custom scene and script
9. Any other ideas ?

BenjaTK:
Closes #341
Fixes #333
Fixes #332
Fixes #328 (At least a bit)
Closes #327

This PR reworks nodes (again) to be created from their script files instead of a separate `.tres` file. This aims to make dynamic slots easier, to reduce the amount of files you have to make per node, and to reduce some ambiguities.

An example for the script of a node is the next one:
```gdscript
@tool
extends GaeaNodeResource
class_name GaeaNodeNoise
## Creates a grid of values from [code]0.0[/code] to [code]1.0[/code] based on a noise algorithm.
##
## Base class for both the 2D and 3D version of this node.


enum NoiseType {
	SIMPLEX = FastNoiseLite.NoiseType.TYPE_SIMPLEX,
	SIMPLEX_SMOOTH = FastNoiseLite.NoiseType.TYPE_SIMPLEX_SMOOTH,
	CELLULAR = FastNoiseLite.NoiseType.TYPE_CELLULAR,
	PERLIN = FastNoiseLite.NoiseType.TYPE_PERLIN,
	VALUE_CUBIC = FastNoiseLite.NoiseType.TYPE_VALUE_CUBIC,
	VALUE = FastNoiseLite.NoiseType.TYPE_VALUE
} # This has to be copied because you can't use FastNoiseLite.NoiseType directly.


func _get_title() -> String:
	return "Noise"


func _get_description() -> String:
	return "Creates a grid of values from [code]0[/bg][/c] to [code]1[/bg][/c] based on a noise algorithm."


func _get_enums_count() -> int:
	return 1


func _get_enum_options(idx: int) -> Dictionary:
	return NoiseType


func _get_enum_default_value(enum_idx: int) -> int:
	return NoiseType.SIMPLEX_SMOOTH


func _get_arguments_list() -> Array[StringName]:
	return [&"frequency", &"lacunarity", &"octaves"]


func _get_argument_type(arg_name: StringName) -> GaeaValue.Type:
	return GaeaValue.Type.INT if arg_name == &"octaves" else GaeaValue.Type.FLOAT


func _get_argument_default_value(arg_name: StringName) -> Variant:
	match arg_name:
		&"frequency": return 0.01
		&"lacunarity": return 2.0
		&"octaves": return 5
	return super(arg_name)


func _get_output_ports_list() -> Array[StringName]:
	return [&"data"]


func _get_output_port_type(output_name: StringName) -> GaeaValue.Type:
	return GaeaValue.Type.DATA


func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) -> Dictionary:
	_log_data(output_port, generator_data)

	var _noise: FastNoiseLite = FastNoiseLite.new()
	_noise.seed = generator_data.generator.seed + salt
	_noise.noise_type = get_enum_selection(0)

	_noise.frequency = _get_arg(&"frequency", area, generator_data)
	_noise.fractal_octaves = _get_arg(&"octaves", area, generator_data)
	_noise.fractal_lacunarity = _get_arg(&"lacunarity", area, generator_data)
	var dictionary: Dictionary[Vector3i, float]
	for x in _get_axis_range(Axis.X, area):
		for y in _get_axis_range(Axis.Y, area):
			for z in _get_axis_range(Axis.Z, area):
				dictionary[Vector3i(x, y, z)] = (_get_noise_value(Vector3i(x, y, z), _noise) + 1.0) / 2.0
	return dictionary


# Hide from the 'Create Node' dialog because this is the base class and only the 2D and 3D versions should show.
func _is_available() -> bool:
	return false


func _get_noise_value(cell: Vector3i, noise: FastNoiseLite) -> float:
	return -1.0

```

This script is also extended by 2 other scripts, `Noise2D` and `Noise3D`. Noise2D, for example, looks like:
```gdscript
@tool
extends GaeaNodeNoise
class_name GaeaNodeNoise2D



func _get_title() -> String:
	return "Noise2D"


func _get_description() -> String:
	return super() + "\n[b]Ignores the z axis.[/b]"


func _get_noise_value(cell: Vector3i, noise: FastNoiseLite) -> float:
	return noise.get_noise_2d(cell.x, cell.y)


func _is_available() -> bool:
	return true

``` 

This then gets made into a node that looks like:
image

This system also allows for dynamically rebuilding the node. If you call `notify_arguments_list_changed` on the `GaeaNodeResource`, it'll tell the graph node to rebuild itself. This is used in the Output node, for example.

Along with all this, this PR adds:
- Enums. They're always at the start of the node and are static, so they won't be rebuilt.
- The Input node, to replace WorldSize and add AreaSize, AreaPosition and AreaEnd.


addons/gaea/graph/components/graph_node_output.gd
```diff
@@ -10,34 +10,38 @@ var resource: GaeaNodeSlotOutput
...
 
 
 ## Sets the corresponding variables.
-func initialize(for_graph_node: GaeaGraphNode, for_idx: int) -> void:
+func initialize(for_graph_node: GaeaGraphNode, for_type: GaeaValue.Type, display_name: String) -> void:
 	graph_node = for_graph_node
-	idx = for_idx
+	type = for_type
+	_label.text = display_name
+	idx = get_index()
```
Zehir, line 23: Maybe I am wrong but I think the initialize method is called from : 
https://github.com/BenjaTK/gaea/blob/f8258c2c7499283b164d5d3ccb2a0ce567a131a6/addons/gaea/graph/nodes/node_slot_output.gd#L16-L20
So it's before the node is in the tree so I don't think the get_index() method will return a correct value

addons/gaea/graph/nodes/node.gd
```diff
@@ -167,29 +242,46 @@ func get_save_data() -> Dictionary:
...
-				if not data.has(child.resource.name):
-					data.set(child.resource.name, child.resource.default_value)
-				if data.get(child.resource.name) != null:
-					child.set_param_value(data[child.resource.name])
+
+	if saved_data.has("enums"):
+		for enum_idx: int in saved_data.get("enums").size():
+			_enum_editors[enum_idx].select(saved_data.get("enums")[enum_idx])
+
+	if saved_data.has("arguments"):
```
Zehir: On the stored file they are called arguments but on the resources they are on the data property ?

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -68,9 +64,256 @@ enum Axis {
...
+## title as in the graph, but can be overriden with [member tree_name_override].
+func get_tree_name() -> String:
+	return tree_name_override if not tree_name_override.is_empty() else _get_title()
+
+## Public version of [method _is_available]. Prefer to override that method over this one.
+func is_available() -> bool:
+	return _is_available()
+
+## Override this method to define the name shown in the title bar of this node.
+## Defining this method is [b]optional[/b], but recommended. If not defined, the description will be empty.
```
Zehir: the title will be "Unnamed"

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -68,9 +64,256 @@ enum Axis {
...
+
+
+## Override this method to add enum properties on the top of the nodes for things like changing
+## operations or types.
+func _get_enums_count() -> int:
+	return 0
+
+
+## Override this method to define the options available for the added enums.[br]
+## Dictionary should be [code]{String: int}[/code]
```
Zehir: Maybe add a note to say that the build-in enum can be used directly ?

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -68,9 +64,256 @@ enum Axis {
...
+func _get_enum_options(idx: int) -> Dictionary:
+	return {}
+
+
+## Override this method if you want to change the display name for the options in the added enums.[br][br]
+## Defining this method is [b]optional[/b]. If not defined, the name will be [code]_get_enum_options(enum_idx).find_key(option_value).capitalize().capitalize()[/code].
```
Zehir: You have 2 .capitalize()

addons/gaea/graph/nodes/node_resource.gd
```diff
@@ -68,9 +64,256 @@ enum Axis {
...
+## Defining this method is [b]optional[/b].
+func _get_enum_default_value(enum_idx: int) -> int:
+	return _get_enum_options(enum_idx).values().front()
+
+
+## Override this method to define the arguments and inputs that will be available in the node.
+## Should be a list of (preferably) [code]snake_case[/code] names.[br][br]
+## Defining this method is [b]required[/b].
+func _get_arguments_list() -> Array[StringName]:
+	return []
```
Zehir: Maybe add a push_warning() call in the base method to display an errors in case the user don't override it

addons/gaea/graph/nodes/output_node_resource.gd
```diff
@@ -11,6 +11,50 @@ class_name GaeaNodeOutput
...
+	return "Output"
+
+
+func _get_arguments_list() -> Array[StringName]:
+	if not is_instance_valid(node) or not is_instance_valid(node.generator):
+		return []
+
+	var layers: Array[StringName]
+	for layer_idx in node.generator.data.layers.size():
+		layers.append(str(layer_idx))
```
Zehir: Here the correct type for the layers is `layers.append(StringName(str(layer_idx)))`
or you can just do `&"%d" % layer_idx` like later in this function

addons/gaea/graph/nodes/root/data/noise/noise.gd
```diff
@@ -0,0 +1,86 @@
...
+	_noise.noise_type = get_enum_selection(0)
+
+	_noise.frequency = _get_arg(&"frequency", area, generator_data)
+	_noise.fractal_octaves = _get_arg(&"octaves", area, generator_data)
+	_noise.fractal_lacunarity = _get_arg(&"lacunarity", area, generator_data)
+	var dictionary: Dictionary[Vector3i, float]
+	for x in _get_axis_range(Axis.X, area):
+		for y in _get_axis_range(Axis.Y, area):
+			for z in _get_axis_range(Axis.Z, area):
+				dictionary[Vector3i(x, y, z)] = (_get_noise_value(Vector3i(x, y, z), _noise) + 1.0) / 2.0
```
Zehir: Why there is + 1.0 / 2.0 ?

Zehir:
Great job, this will make the node creation way easier

addons/gaea/graph/nodes/node.gd
BenjaTK: Yeah I still have to change that other name

addons/gaea/graph/nodes/root/data/noise/noise.gd
BenjaTK: Because noises in Godot return values from -1.0 to 1.0 but it's more common to use 0.0-1.0 (it works nicer with colors, for example)

Zehir: ah OK did not know that. Also for performance it's maybe better to multiply by 0.5 instead because multiply is faster than divide

addons/gaea/graph/components/graph_node_output.gd
BenjaTK: This has now changed and is called from `_add_output_slot`. The GaeaNodeSlotOutput resource will be removed

addons/gaea/graph/nodes/node.gd
BenjaTK: Done

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
```diff
@@ -18,10 +20,12 @@ offset_right = 211.0
 offset_bottom = 43.0
 script = ExtResource("1_xfmar")
 
-[node name="Label" type="Label" parent="."]
+[node name="Label" type="RichTextLabel" parent="."]
```
Zehir, line 23: Why a RichTextLabel if bbcode is not enabled?

addons/gaea/graph/graph_nodes/generic_classes/number_operation.gd
```diff
@@ -0,0 +1,197 @@
...
+	var output: String
+	var conversion: Callable
+	func _init(_args: Array[StringName], _output: String, _conversion: Callable):
+		args = _args
+		output = _output
+		conversion = _conversion
+
+
+## All possible operations.
+var OPERATION_DEFINITIONS: Dictionary[Operation, Definition] : get = _get_operation_definitions
```
Zehir, line 42: I think we can make this Dictionary static

addons/gaea/graph/graph_nodes/generic_classes/number_operation.gd
```diff
@@ -0,0 +1,197 @@
...
+	for arg_name: StringName in OPERATION_DEFINITIONS[operation].args:
+		args.append(_get_arg(arg_name, _area, generator_data))
+	return _get_new_value(operation, args)
+
+
+func _get_new_value(operation: Operation, args: Array) -> Variant:
+	return OPERATION_DEFINITIONS[operation].conversion.callv(args)
+
+
+func _get_operation_definitions() -> Dictionary[Operation, Definition]:
```
Zehir, line 145: With this function static too

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
```diff
@@ -0,0 +1,128 @@
...
+	var output: String
+	var conversion: Callable
+	func _init(_args: Array[StringName], _output: String, _conversion: Callable):
+		args = _args
+		output = _output
+		conversion = _conversion
+
+
+## All possible operations.
+var OPERATION_DEFINITIONS: Dictionary[Operation, Definition] : get = _get_operation_definitions
```
Zehir, line 25: This one too can be static I think

addons/gaea/graph/graph_nodes/root/data/operations/scalar/data_operation.gd
```diff
@@ -0,0 +1,70 @@
...
+
+func _get_operation_definitions() -> Dictionary[Operation, Definition]:
+	var definitions := super()
+	for definition: Definition in definitions.values():
+		# Kind of horrible code but it's fine
+		definition.output = definition.output.replace("a ", "A ")
+		definition.output = definition.output.replace(" a", " A")
+		definition.output = definition.output.replace("a,", "A,")
+		definition.output = definition.output.replace("(a)", "(A)")
+	definitions.erase(Operation.Power)
```
Zehir, line 52: This will not erase the parent definition if you did not duplicate it ?

Zehir (approved):
Look good to me, A fiew comment but they can be handled later. Yolo merge !

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
BenjaTK: If someone wants they can use bbcode in the display name method

addons/gaea/graph/graph_nodes/generic_classes/number_operation.gd
BenjaTK: The thing is it's different in IntOp and DataOp since they remove some entries

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
Zehir: But it's not enabled in the Control

addons/gaea/graph/graph_nodes/generic_classes/number_operation.gd
Zehir: Then we will need to move the var in child classes

cullumi (dismissed):
What I'm noticing first off is that in order for a custom class to appear in the Search Help menu so you can view it's documentation, you need to place it's initial description *above* the "extends" line.

I believe all of Gaea's code does not satisfy this requirement, and is therefore absent from the Search Help menu.

I wouldn't know if this were intentional, though; could technically be saved for a separate PR, as well.

BenjaTK:
> What I'm noticing first off is that in order for a custom class to appear in the Search Help menu so you can view it's documentation, you need to place it's initial description *above* the "extends" line.
> 
> I believe all of Gaea's code does not satisfy this requirement, and is therefore absent from the Search Help menu.
> 
> I wouldn't know if this were intentional, though; could technically be saved for a separate PR, as well.

I think you're experiencing a bug where Godot doesn't load documentation properly. The comment docs work fine when you update them until you restart Godot, then they disappear. It's been fixed in 4.5 tho

addons/gaea/graph/graph_nodes/root/conditional/constants/bool_constant.gd
```diff
@@ -0,0 +1,16 @@
+@tool
+class_name GaeaNodeBoolConstant
+extends GaeaNodeConstant
+## [bool] constant.
+
+
+func _get_output_port_type(_output_name: StringName) -> GaeaValue.Type:
+	return GaeaValue.Type.BOOLEAN
```
cullumi, line 8: Using GaeaNodeBoolConstant as an example, I'm noticing that changing a node in code (such as changing it's output port type to `VECTOR3` from `BOOLEAN`), then hitting "Reload create node list" does not change already created nodes. These nodes get reloaded properly after deselecting the GaeaGenerator and reselecting it again, however.

But who knows, this may be a topic worthy of it's own feature PR.

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
BenjaTK: Oops

cullumi (approved):
Leaving my reloading commentary for a separate QOL PR. I'm satisfied with calling this PR complete. I like the folder reorganization, and the new approach to defining sub-nodes.

PR #353, opened 2025-04-27
Rework data casting functionality

Zehir:
Refactor the data casting logic by introducing a new class for handling casting methods.


Zehir:
This allow no to do something like this ;
image

I did not include the scene update in this PR because it will conflicts with #330

I used a Vector2 operation to set the Z component to 0 on the last casting but maybe the node is ignoring it, I did not check the code.

PR #359, opened 2025-04-27
Rework `GaeaMaterial`s to allow them to be sampled and make `GaeaMaterialGradient` into a material

### LINKED ISSUE #343: Unify GaeaMaterial and GaeaGradientMaterial with Recursive Sampling
# Feature Request: Unify GaeaMaterial and GaeaGradientMaterial with Recursive Sampling

## Description

The idea is to merge `GaeaGradientMaterial` into `GaeaMaterial` to simplify the material system.

The new `GaeaMaterial` will include:
- A virtual method `_is_sampled() -> bool`.
- A method `sample(value: float) -> GaeaMaterial`.
- The existing `get_resource() -> GaeaMaterial`, which typically returns `self`.

## Behavior

Mappers will:

1. Check if the material is sampled using `_is_sampled()`.
2. If true, recursively call `sample(value)` with the cell value until `_is_sampled()` returns false.
3. Finally, call `get_resource()` on the resolved material and store it in the map.

This removes the need for a dedicated gradient mapper. All mappers will handle sampled materials the same way.

For example, `threshold_mapper` can still use the value to determine if a cell should have a material, and then use `sample()` to get the actual material if the material is sampled.

On the wire we will have only the red `GaeaMaterial` class, which will be a base class for all materials. The blue `GaeaGradientMaterial` will be removed.

## Question

About the `sample()` method, should the value be remapped to the filtered range that go throught the `_passes_mapping` method ?

The workaround for that would be to use the `GaeaNodeFlagsFilter` instead of the `GaeaNodeFlagsMapper` and have a `NormalizeData` Node that will remap the values between 0 and 1 using the current min/max values of the data. FOr the vector the `normalize` method just set a lenght of 1, maybe we should find a better name for that. Maybe just a remap node that have only target values and use min/max values.

In the end the `GaeaNodeFlagsMapper` is just a `GaeaNodeFlagsFilter` with a `GaeaNodeMapper` node smooshed in.

Maybe we should extend the changes and just have one mapper and more data operation nodes. So we don't have multiples nodes that do the same thing.

## Benefits

- Simplifies the class hierarchy and removes special cases.
- Allows different kinds of random materials: 
  - Global (one material for the entire map) by using only `get_resource()` method.
  - Per-cell (like the current `RandomMaterial`) by using `sample()` method.
- Provides a unified, extensible system for future sampling-based materials.

Zehir:
# Overview

This is another rework, but a smaller one this time. The goal is to simplify Materials

Closes #343.

## Changes

- Consolidated to a single base material that can now be either sampled or not.
- Adopted the same design pattern as the nodes, using virtual methods combined with public methods.
- **Removed** `GradientMapper`, as it was no longer used; it has been replaced by `BasicMapper` (with migration script).
- Moved materials into a subdirectory to better organize and clarify their responsibilities.
- Renamed `RandomMaterial` into `GaeaPointwiseRandomMaterial` who pick one random material for each cell.
- Added `Gaea` prefix for all material classes
- Added `GaeaPickRandomMaterial` who pick just one random material and will use it for the complete mapper.
- Added `define_rng(generator_data: GaeaData) -> RandomNumberGenerator` on the resource that define the current seed with the node salt and return a `RandomNumberGenerator` instance to be used for the mapper. Now the random material mapper return the same result with the same seed.

## Notes

Unfortunately, this PR currently conflicts with #356. So for now, it includes the full commit history, but it effectively uses only the commit d1e8c66ed9f28237d63c164e2fe16e607181f7f7, I will cherry pick it when #356 is merged but it's ready for review


cullumi:
#356 has now been merged

Zehir:
@cullumi ready for review

addons/gaea/resources/materials/distribution/pointwise_random_material.gd
```diff
@@ -0,0 +1,24 @@
...
+## # 70% chance for grass, 30% for stone
+## [/codeblock]
+
+
+func _is_sampled() -> bool:
+	return true
+
+
+## Return the random picked material.
+func get_sampled_resource(rng: RandomNumberGenerator, _value: float) -> GaeaMaterial:
```
BenjaTK: This should be the private version of the function (and also shouldn't have the doc comment)

BenjaTK (changes requested):
I like this, tho it might be a bit hard to remember which methods have to be overriden for every type of material.

cullumi (changes requested):
I'm liking everything I'm seeing, it's nice to see a consistent approach across the codebase.

Minor quirks are what I noticed, at most.

addons/gaea/resources/materials/distribution/gradient_material.gd
```diff
@@ -60,7 +60,8 @@ func _sort_points() -> void:
 	)
 	points_sorted.emit()
 
-func sample(value: float) -> GaeaMaterial:
+
+func get_sampled_resource(rng: RandomNumberGenerator, value: float) -> GaeaMaterial:
```
cullumi: Should use the private method here

addons/gaea/resources/materials/distribution/pick_random_material.gd
```diff
@@ -42,8 +41,8 @@ var _random = RandomNumberGenerator.new()
 
 
 ## Return the random picked material.
-func get_resource() -> GaeaMaterial:
-	var material_index: int = _random.rand_weighted(weights)
+func get_resource(rng: RandomNumberGenerator) -> GaeaMaterial:
```
cullumi: Another place where the private method maybe should be used?

cullumi:
I noticed there were spots where you put some code to catch recursive loops. Seems like a smart idea to me! 👍

addons/gaea/resources/materials/distribution/pointwise_random_material.gd
Zehir: That ont is a virtual method that need to be overriden when you have a sampled material. I am not sure what you want me to do

Zehir: Done, I have updated the other one too

Zehir:
@cullumi I can't see what changes you requested
@BenjaTK I have updated the PR

addons/gaea/resources/materials/distribution/gradient_material.gd
Zehir: Already updated in previous commit

addons/gaea/resources/materials/distribution/pick_random_material.gd
Zehir: Updated in latest commit

addons/gaea/assets/types/sampled_material.svg
BenjaTK, line 1: I really don't think this icon makes sense for a sampled material. The color doesn't match and it's just the gradient icon. I think just having it be the same icon for all materials is fine. Also, the PickRandomMaterial uses it when it's not sampled.

addons/gaea/resources/materials/distribution/gradient_material.gd
BenjaTK, line 1: `_init` should also call `super()`.

addons/gaea/graph/graph_nodes/root/map/mappers/basic_mapper.gd
```diff
@@ -36,14 +36,23 @@ func _get_required_arguments() -> Array[StringName]:
 func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) -> Dictionary:
 	_log_data(output_port, generator_data)
 
-	var grid_data = _get_arg(&"reference_data", area, generator_data)
-	var material: GaeaMaterial = _get_arg(&"material", area, generator_data)
+	var grid: Dictionary[Vector3i, GaeaMaterial] = {}
+	var grid_data := _get_arg(&"reference_data", area, generator_data) as Dictionary
+	var material := _get_arg(&"material", area, generator_data) as GaeaMaterial
+	var rng = define_rng(generator_data)
```
BenjaTK: Missing type declaration for var

addons/gaea/graph/graph_nodes/root/map/placing/rules_placer.gd
```diff
@@ -56,11 +56,16 @@ func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) ->
 
 	var grid_data: Dictionary = _get_arg(&"reference_data", area, generator_data)
 	var material: GaeaMaterial = _get_arg(&"material", area, generator_data)
-
+	var rng = define_rng(generator_data)
```
BenjaTK: Here aswell

addons/gaea/graph/graph_nodes/root/map/placing/random_scatter.gd
```diff
@@ -41,15 +41,19 @@ func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) ->
 
 	var grid_data: Dictionary = _get_arg(&"reference_data", area, generator_data)
 	var material: GaeaMaterial = _get_arg(&"material", area, generator_data)
-
-	seed(generator_data.generator.seed + salt)
+	var rng = define_rng(generator_data)
```
BenjaTK: Missing type declaration for var

addons/gaea/resources/materials/data/tilemap_material.gd
```diff
@@ -1,5 +1,6 @@
 @tool
-class_name TileMapMaterial
+@icon("../../../assets/types/material.svg")
```
BenjaTK: If the base class already has the icon set this doesn't need it.

BenjaTK (changes requested):
I think the names should be changed to follow the same format as the renderers: `TileMapGaeaMaterial`, `PointwiseRandomGaeaMaterial`.

BenjaTK (approved):
LGTM

PR #360, opened 2025-04-28
Removed uid prefix in migration

Zehir:
For some reason in my game with the addon Godot 4.5 seems to resolve the uid in the string in the migration script even if that make no sense.

This PR will collide with previous PR, will cherry pick once the previous PR are merged


BenjaTK:
Does this still work in 4.4 though?

Zehir:
Yes it's just a string that is changed for the migration only, it does not change the saved value

Zehir:
This one will collide with #359 so lets do #359 first

Zehir:
This PR has been rebased and is now ready for review

cullumi (approved):
Seems pretty sane to me!

PR #367, opened 2025-04-29
Add new shapes to `GaeaNodeFalloff`

Zehir:
# Rework falloff node with new shapes
## Description

This PR is a rework of the falloff node with some new features and bug fix.

Because an image is still better then a lot of words, here the old node and the new node. In the screenshot beflow the FalloffMap2 is the new version of the FalloffMap.

There is now 4 shapes included with the node. I did not include each variation in the add node tree, just the basic FalloffMap node like before.

- Square, like before just a square
- Rounded Square, a square but with round corner
- Circle, just a circle
- Squircle, a circle in the middle but a rounded square on the border.

## Sample images
image
image
image

Works also with lower "resolution", the old one had an offset issue that is now fixed;
image

It's also support non square shapes (I used the PR #366 for the preview);

image
image

## Q&A
- Why there is that many var defined in the FalloffSampler class ?
- It's because of performances. It's some calculation that need to be do only once and not for each cell.

### Performance tests
I tried to generate a big grid and it's taking in average 530 ms to generate the data.
I tried without all var it's taking 550 ms in average
I also tried with a PackedFloat32Array but was worse: 580 ms

For a standard grid of 64*64 it's taking 4 ms

## References
https://old.reddit.com/r/proceduralgeneration/comments/kaen7h/new_video_on_procedural_island_noise_generation/gfjmgen/
https://www.redblobgames.com/maps/terrain-from-noise/#islands

### DIFF
--- a/addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
+++ b/addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
@@ -8,6 +8,121 @@ class_name GaeaNodeFalloffMap
 ## For lower [param end] values, the generated 'square' will be smaller.[br]
 ## Multiplying this with a [GaeaNodeSimplexSmooth]'s generation can create island-looking terrains.
 
+enum FalloffShape {
+	Square,
+	RoundedSquare,
+	Circle,
+	Squircle,
+}
+
+
+class FalloffSampler:
+	var area: AABB
+	var start: float
+	var end: float
+	var pos_x: float
+	var pos_y: float
+	var size_x: float
+	var size_y: float
+	var size_x_is_smaller: bool
+	var size_y_is_smaller: bool
+	var size_x_ajusted: float
+	var size_y_ajusted: float
+	var size_x_half: float
+	var size_y_half: float
+	var size_x_minus_y_half: float
+	var size_y_minus_x_half: float
+
+	func _init(_area: AABB, _start: float, _end: float):
+		area = _area
+		start = _start
+		end = _end
+		pos_x = area.position.x
+		pos_y = area.position.y
+		size_x = area.size.x
+		size_y = area.size.y
+		var ratio: float = 1.0
+		size_x_is_smaller = size_x <= size_y
+		if not size_x_is_smaller:
+			ratio = float(size_y) / float(size_x)
+		size_y_is_smaller = size_y <= size_x
+		if not size_y_is_smaller:
+			ratio = float(size_x) / float(size_y)
+		size_x_ajusted = size_x * ratio
+		size_y_ajusted = size_y * ratio
+		size_x_half = size_x * 0.5
+		size_y_half = size_y * 0.5
+		size_x_minus_y_half = size_x - size_y_half
+		size_y_minus_x_half = size_y - size_x_half
+		_on_init()
+
+	func _on_init():
+		pass
+
+	func cx(x: float) -> float:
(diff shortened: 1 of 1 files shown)


BenjaTK:
The squircle should probably be removed, rounded square does almost entirely the same

Zehir:
> The squircle should probably be removed, rounded square does almost entirely the same

I would prefer keep it because it's not exactly the same and it's just 3 more lines of code

cullumi:
> > The squircle should probably be removed, rounded square does almost entirely the same
> 
> I would prefer keep it because it's not exactly the same and it's just 3 more lines of code

There's probably no real downsides to holding onto, yeah. While the differences are subtle, they are technically different.

cullumi (approved):

addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
```diff
@@ -8,6 +8,121 @@ class_name GaeaNodeFalloffMap
 ## For lower [param end] values, the generated 'square' will be smaller.[br]
 ## Multiplying this with a [GaeaNodeSimplexSmooth]'s generation can create island-looking terrains.
 
+enum FalloffShape {
+	Square,
```
BenjaTK: enum names should be UPPERCASE

BenjaTK: This is also something that needs to be changed in the operation nodes

addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
```diff
@@ -8,6 +8,121 @@ class_name GaeaNodeFalloffMap
...
+		size_x_half = size_x * 0.5
+		size_y_half = size_y * 0.5
+		size_x_minus_y_half = size_x - size_y_half
+		size_y_minus_x_half = size_y - size_x_half
+		_on_init()
+
+	func _on_init():
+		pass
+
+	func cx(x: float) -> float:
```
BenjaTK: These method names should be more descriptive.

BenjaTK (changes requested):

addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
Zehir: Made a PR for that https://github.com/gaea-godot/gaea/pull/372

addons/gaea/graph/graph_nodes/root/data/generation/falloff_map.gd
```diff
@@ -8,6 +8,121 @@ class_name GaeaNodeFalloffMap
 ## For lower [param end] values, the generated 'square' will be smaller.[br]
 ## Multiplying this with a [GaeaNodeSimplexSmooth]'s generation can create island-looking terrains.
 
+enum FalloffShape {
+	SQUARE,
+	ROUNDEDSQUARE,
```
BenjaTK: This should be ROUNDED_SQUARE.

BenjaTK (changes requested):
Missing hints for `start` and `end` to be withing 0.0-1.0.

Zehir:
> Missing hints for `start` and `end` to be withing 0.0-1.0.

They should not be clamped to 0.0-1.0, you may want use values outside the "standard" range
image

Issue #260, opened 2025-04-08
Switch to using UIDs for resource paths


cullumi:
**Is your feature request related to a problem? Please describe.**
Resource paths (`res://`) break when users choose to place the plugin in a folder other than `res://addons`.

**Describe the solution you'd like**
To solve these issues, we should use the new UIDs that got expanded in Godot 4.4: https://godotengine.org/article/uid-changes-coming-to-godot-4-4/

**Describe alternatives you've considered**
Relative paths (i.e. `../../assets/types`) are the current alternative, but break when reorganizing Gaea's directories.

**Is there a reason this should be core and not a project-specific feature?**
This is fundamental to Gaea working as expected out-of-the-box.

**Additional context**
N/A

cullumi:
I've already completed most of this work, but need to investigate an error I've been encountering to confirm whether I created it. Will have a PR soon.

PR #370, opened 2025-05-01
Make categories collapsable

BenjaTK:
image

image

The slots still showing wasn't intended at first but it makes a lot of sense imo and is just better. The state is saved as if it was the value of the argument.

This PR also:
- Makes category text bold.
- Changes the root of the base `GaeaArgumentEditor` to a `BoxContainer` so it can be either vertical or horizontal (this was previously done by just breaking the inheritance, like in the range one).

## To-do for this PR:

- [x] Add a hint to hide the collapse button.


addons/gaea/graph/graph_nodes/node.gd
```diff
@@ -3,7 +3,7 @@ class_name GaeaGraphNode
 extends GraphNode
 ## The in-editor representation of a [GaeaNodeResource] to be used in the Gaea bottom panel.
 
-const _PreviewTexture = preload("uid://dns7s4v8lom4t")
+const _PreviewTexture = preload("res://addons/gaea/graph/components/preview_texture.gd")
```
Zehir, line 6: Why you removed the UID here ? It's was one step for #260

Zehir (approved):
LGTM

addons/gaea/graph/graph_nodes/node.gd
BenjaTK: oops sorry. Godot was giving me problems with that one, said it was missing. forgot to not add that chsnge

Zehir: > oops sorry. Godot was giving me problems with that one, said it was missing. forgot to not add that chsnge

Did you had missing uid "uid://a" too ? Sometime I get them but I don't know from where this errors is comming from

BenjaTK: no just this one, but it didn't even want to load the script. the whole addon was broken

Zehir:
I like the fact where the slot are smooshed toogether but I don't know, maybe we should just hide them and remove the separator bellow except if the slot is connected and in that case don't hide the slot.

Or maybe have a state where the node is selected and not selected where when selected the node auto show the the fields but you may have issue connecting wire if you did not hide the slot yourself

Or maybe a checkbox like the 4.5 inspector ? so the true/false would be used in code to skip sections of the code

BenjaTK:
The checkbox in 4.5 is to disable/enable the features inside the categories, which isn't used here. The reason I like that the slots still show is that any input wires still show up and you can tell how many arguments you've hidden. Plus I like how it looks, lol

BenjaTK:
Plus if we hide the slots we have to do a whole thing with connection indexes and it's a mess just for a simple feature

Zehir:
> The checkbox in 4.5 is to disable/enable the features inside the categories, which isn't used here. The reason I like that the slots still show is that any input wires still show up and you can tell how many arguments you've hidden. Plus I like how it looks, lol

the idea of enable/disable was for my test with the noise node ;
Where domain_wrap_enabled is a real enable/disable the domainwrap category
image

BenjaTK:
Couldn't that just be a boolean property instead of an enum?

Zehir:
> Couldn't that just be a boolean property instead of an enum?

yes but I wanted to hide the slots when it's was disabled but collapsing it is a good solution too

Zehir:
Do we have default state for the collapsed ?

BenjaTK:
> Do we have default state for the collapsed ?

yeah, it's expanded. I like it that way because it means all properties are visible when you first add the node

Zehir:
Is it saved in the graph file ? I did not see that part of the code

BenjaTK:
> Is it saved in the graph file ? I did not see that part of the code

Saved as if it was the argument's value (since categories are arguments and we already have the argument saving). It's a true/false (collapsed or not).

BenjaTK:
Forgot to make the collapsable hint disable pressing in the category label, will do it later today.

addons/gaea/graph/components/argument_editors/category.gd
```diff
@@ -0,0 +1,66 @@
+@tool
```
Zehir, line 1: The uid file is missing (category.gd.uid)

addons/gaea/graph/components/argument_editors/category.gd
```diff
@@ -0,0 +1,66 @@
...
+class_name GaeaArgumentCategory
```
Zehir, line 2: Some UID link seems broken, I got this godot warning
  WARNING: (2) scene/resources/resource_format_text.cpp:447 - res://addons/gaea/graph/components/argument_editors/category.tscn:3 - ext_resource, invalid UID: uid://b67wud5egfs8o - using text path instead: res://addons/gaea/graph/components/argument_editors/category.gd

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
```diff
@@ -1,45 +1,35 @@
-[gd_scene load_steps=6 format=3 uid="uid://dy3oumbnydlmp"]
+[gd_scene load_steps=6 format=3 uid="uid://t4osuglcgg6l"]
```
Zehir, line 1: The range slider seems broken, I can't slide it

addons/gaea/graph/graph_nodes/node.gd
Zehir: This was marked as resolved but the uid is still missing

Zehir (changes requested):

addons/gaea/graph/components/argument_editors/range_argument_editor.tscn
BenjaTK: Fixed

cullumi (approved):
I like the concept, and I think if the collapsed slot connections are weird we can pretty reasonable take a different approach later for them.

PR #371, opened 2025-05-02
Add lerp to `DatasOp`

BenjaTK:


BenjaTK:
@Zehir

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
```diff
@@ -122,6 +138,8 @@ func _get_operation_definitions() -> Dictionary[Operation, Definition]:
 		Operation.Multiply:
 			Definition.new([&"a", &"b"], "A * B", func(a: Variant, b: Variant): return a * b),
 		Operation.Divide:
-			Definition.new([&"a", &"b"], "A / B", func(a: Variant, b: Variant): return 0 if is_zero_approx(b) else a / b)
+			Definition.new([&"a", &"b"], "A / B", func(a: Variant, b: Variant): return 0 if is_zero_approx(b) else a / b),
+		Operation.Lerp:
+			Definition.new([&"a", &"b", &"weight"], "lerp(a, b, weight)", lerp)
```
Zehir: Should we use `lerpf` instead of `lerp` ?

Zehir (approved):
LGTM, just one comment

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
```diff
@@ -75,9 +78,16 @@ func _get_arguments_list() -> Array[StringName]:
...
+func _get_argument_type(arg_name: StringName) -> GaeaValue.Type:
+	if arg_name == &"weight":
+		return GaeaValue.Type.FLOAT
 	return GaeaValue.Type.DATA
-
+	
+	
+func _get_argument_hint(arg_name: StringName) -> Dictionary[String, Variant]:
+	if arg_name == &"weight":
+		return {"min": 0.0, "max": 1.0}
```
Zehir, line 89: I am starting to think we could need a NORMALIZED_FLOAT type with a slider of a range 0.0 to 1.0, we starting to have more and more node with a float that is expected to be between 0 and 1. But will still use the number slot kind, the gray circle

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
```diff
@@ -42,6 +43,8 @@ func _get_description() -> String:
 			return "Multiplies all cells in [param B] with all cells in [param A]."
 		Operation.Divide:
 			return "Adds all cells in [param A] by all cells in [param B]."
+		Operation.Lerp:
```
Zehir: Do we want to have a single enum somewhere and reference it everywhere ? And use a filter using OPERATION_DEFINITIONS dictionary for the _get_enum_options function

BenjaTK: nah I think it's okay, it's only 2 enums after all. having it global would make it so it depends on another script

Zehir: Ok good point too

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
BenjaTK: I don't think this is necessary tbh, just setting up the hints is fine.

Zehir: Maybe add a slider to the float editor if we have a min/max hint but for an other time ?

addons/gaea/graph/graph_nodes/root/data/operations/datas/datas_operation.gd
BenjaTK: Yeah fair point

PR #373, opened 2025-05-03
Add a Texture data type and a TextureSampler node

BenjaTK:
This PR adds a Texture data type, colored orange and with the diamond shape previously used in the Gradient type (which has been removed in #359). 

The `Texture2DParameter` and `Texture3DParameter` nodes can be input into the `TextureSampler` node to output 4 data grids for each of the 4 image channels (r, g, b, a).

This image:
image

Gets sampled into:
Peek 2025-05-02 22-32
Edit: `apply_seed` has been removed.

~~If `apply_seed` is on, and the texture is a `NoiseTexture2D/3D`, it'll set its seed to the current generation seed (+ node's salt).~~


Zehir:
What happen if the texture size is not the same size as the map size ? You will lerp to get a value ? Resize it if it's a noise texture ?

BenjaTK:
> What happen if the texture size is not the same size as the map size ? You will lerp to get a value ? Resize it if it's a noise texture ?

Nothing. If the texture is smaller it just doesn't go over the empty pixels and if it's bigger it just cuts off.

BenjaTK:
Did not mean to mark it as ready for review and I can't seem to revert it in mobile

BenjaTK:
image

Used the image icon for the texture parameter. I know it's not exactly the Texture icon but the base Texture class doesn't have one and this is the one that fits the best. I decided not to color it orange and use the base color instead because 1. it's easier to get the icon directly from the editor and 2. the color can be changed in the settings anyways so if you did that it wouldn't match.

addons/gaea/resources/gaea_value.gd
```diff
@@ -249,59 +256,3 @@ static func get_editor_for_type(for_type: GaeaValue.Type) -> PackedScene:
 		GaeaValue.Type.RULES:
 			return preload("uid://dy4n2a5hkaxsb")
 	return preload("uid://i2nwlab8rau")
-
-
-## Get property type hint, this is mostly used with [method Object._validate_property].
```
Zehir, line 244: Not sure to understand why this was removed ?

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -0,0 +1,78 @@
...
+
+func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) -> Dictionary:
+	_log_data(output_port, generator_data)
+	var texture: Texture = _get_arg(&"texture", area, generator_data)
+	if not is_instance_valid(texture):
+		return {}
+		
+	if _get_arg(&"apply_seed", area, generator_data):
+		if texture is NoiseTexture2D or texture is NoiseTexture3D:
+			texture.noise.seed = generator_data.generator.seed + salt
```
Zehir: The problem with that is that you changed the seed property on the external resource instead of locally. I think you need to duplicate the resource first
image

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -0,0 +1,78 @@
+@tool
```
Zehir, line 1: I tried to play with the texture settings and adding a Color Ramp to it frooze Godot for 2 seconds on generation;
image

Same problem with the "As Normal Map" toggled on

But maybe it's just because it's take longer to generate the texture, need to be checked

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -0,0 +1,78 @@
...
+# List of all the outputs, preferably in &"snake_case"
+func _get_output_ports_list() -> Array[StringName]:
+	return [&"r", &"g", &"b", &"a"]
+	
+	
+func _get_output_port_display_name(output_name: StringName) -> String:
+	return output_name
+
+
+func _get_output_port_type(output_name: StringName) -> GaeaValue.Type:
```
Zehir: Godot warning : The parameter "output_name" is never used in the function "_get_output_port_type()". If this is intended, prefix it with an underscore: "_output_name".

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -0,0 +1,78 @@
...
+			texture.noise.seed = generator_data.generator.seed + salt
+			
+	var r_grid: Dictionary
+	var g_grid: Dictionary
+	var b_grid: Dictionary
+	var alpha_grid: Dictionary
+	
+	var slices: Array[Image] # Only one is texture is 2D
+	if texture is Texture2D:
+		slices = [texture.get_image()]
```
Zehir, line 50: It's seems that `texture.get_image()` return null with the `CurveTexture` because it's seems to be invalid.
`  ERROR: res://addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd:64 - Cannot call method 'get_used_rect' on a null value.`
Same with the `PlaceholderTexture2D` or `CurveXYZTexture` so I suppose just a null check would be enought

I also tried with a `GradientTexture1D` and I just get a line so I think it's expected

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -0,0 +1,78 @@
...
+
+
+func _get_data(output_port: StringName, area: AABB, generator_data: GaeaData) -> Dictionary:
+	_log_data(output_port, generator_data)
+	var texture: Texture = _get_arg(&"texture", area, generator_data)
+	if not is_instance_valid(texture):
+		return {}
+		
+	if _get_arg(&"apply_seed", area, generator_data):
+		if texture is NoiseTexture2D or texture is NoiseTexture3D:
```
Zehir: If the texture is a `SVGTexture` maybe we could change the texture size to match the cell size, or maybe it's the user responsibility to do it ?

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
Zehir: The NoiseTexture3D or 2D may not have the noise object set;
  ERROR: res://addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd:45 - Invalid assignment of property or key 'seed' with value of type 'int' on a base object of type 'null instance'.
image

addons/gaea/resources/gaea_value.gd
```diff
@@ -27,6 +27,7 @@ enum Type {
 	RANGE = 100,
 	MATERIAL = 101, ## A [GaeaMaterial].
 	GRADIENT = 102, ## A [GaeaMaterialGradient].
+	TEXTURE = 103, ## A [Texture].
```
Zehir: Do you want to re use the 102 once the GRADIENT is removed ?

addons/gaea/resources/gaea_value.gd
```diff
@@ -158,8 +161,8 @@ static func get_default_color(type: Type) -> Color:
 		Type.MAP:
 			return Color("27ae60") # GREEN
 		# Reserved for later use
-		#SlotType.TEXTURE: # ORANGE
-		#	return Color("e67e22")
+		Type.TEXTURE: # ORANGE
```
Zehir: Could you please put the Orange comment after the colors like the other ones, it's tickling my brain ;)

Zehir (commented):
Looking good but some changed needed

addons/gaea/resources/gaea_value.gd
BenjaTK: This was used for the exports in the GaeaNodeSlot resource and it's no longer used or needed

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
BenjaTK: That should be something the user does, not us

addons/gaea/resources/gaea_value.gd
Zehir: Ah ok it's not related to the texture sampler np

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
Zehir: Am not sure about that because with the threading if we generate multiple area at the same time with the same reference of the resource we may get unexpected results

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
BenjaTK: Pretty sure these are all Godot problems and there's no way for us to fix this. If anyone has any ideas I'd appreciate them but if not we can look out for it in the future ig

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
BenjaTK: I'd say it's their responsibility.

Zehir (approved):
LGTM

PR #379, opened 2025-05-05
Rebuild graph nodes when their resource script is changed

### LINKED ISSUE #348: Reload graph nodes when their script is changed
Right know if you make changes to the node's script, like changing argument list, it'll only show the changes if `notify_argument_list_changed` is called or if the user deselects and selects the node (essentially reloading the graph).

Could probably listen to changes made to the script and call `_rebuild()` on the GraphNode.

Zehir:
## Description

Closes #348 

This PR ensures that **graph nodes**  automatically rebuild when their associated resource script is updated.

The reload only occurs if the script was successfully compiled, preventing issues from broken or invalid scripts.

There is a small delay after save but I suppose it's the **compilation** time.

Godot_v4 5-dev3_win64_6LomMkjCcV

On script compilation error there is no added error to tell that the compilation failed and the node could not be reloaded.

image

### DIFF
--- a/addons/gaea/editor/graph_edit.gd
+++ b/addons/gaea/editor/graph_edit.gd
@@ -16,6 +16,7 @@ func _ready() -> void:
 	if is_part_of_edited_scene():
 		return
 	add_theme_color_override(&"connection_rim_color", Color("141414"))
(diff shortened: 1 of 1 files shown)


BenjaTK:
Love this, LGTM

PR #380, opened 2025-05-05
Validate Script Type Before Adding to Node Tree

### LINKED ISSUE #362: Add a guard clause to make sure the script is a node script
In the create_node_tree we should make sure that the script is a node_resource and ignore any script that is not that. In the main main list of node it's not a problem because we don't put not node script in the root directory but for the custom node path it's can have anything in it

  ERROR: res://addons/gaea/editor/create_node_tree.gd:86 - Trying to assign value of type 'tile_set_material_resource.gd' to a variable of type 'node_resource.gd'.

Zehir:
## Description

Closes #362

This PR ensures that only **valid scripts** (i.e., `GaeaNodeResource` scripts) are added to the tree during creation. Which can happen if users place unrelated script files in the root or custom root directories.

An alternative approach considered was to instantiate the object and check its type at runtime. However, this would involve creating instances of arbitrary scripts, which could lead to unintended side effects—even if the implementation would require less code.

### DIFF
--- a/addons/gaea/editor/create_node_tree.gd
+++ b/addons/gaea/editor/create_node_tree.gd
@@ -83,13 +83,22 @@ func _populate_dict_with_files(folder_path: String, dict: Dictionary) -> Diction
 
 
 		if file_name.ends_with(".gd"):
-			var resource: GaeaNodeResource = load(file_path).new()
-			if resource is GaeaNodeResource:
-				if resource.is_available():
-					var sub_idx: int = 0
-					for item in resource.get_tree_items():
-						sub_idx += 1
-						dict.get_or_add(item.get_tree_name() + str(idx) + str(sub_idx), item)
+			var script := load(file_path)
+			if script is GDScript:
+				var is_valid_node_resource := false
+				var base_script: GDScript = script
+				while is_instance_valid(base_script):
+					base_script = base_script.get_base_script()
+					if base_script == GaeaNodeResource:
+						is_valid_node_resource = true
+						break
+				if is_valid_node_resource:
+					var resource: GaeaNodeResource = script.new()
+					if resource.is_available():
+						var sub_idx: int = 0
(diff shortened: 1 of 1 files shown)


addons/gaea/editor/create_node_tree.gd
```diff
@@ -83,13 +83,22 @@ func _populate_dict_with_files(folder_path: String, dict: Dictionary) -> Diction
...
-				if resource.is_available():
-					var sub_idx: int = 0
-					for item in resource.get_tree_items():
-						sub_idx += 1
-						dict.get_or_add(item.get_tree_name() + str(idx) + str(sub_idx), item)
+			var script := load(file_path)
+			if script is GDScript:
+				var is_valid_node_resource := false
+				var base_script: GDScript = script
+				while is_instance_valid(base_script):
```
BenjaTK, line 90: Huh, is there really no better way to do this?

Zehir: Maybe the is keyword could work, I will test tomorrow. Because today I notice that the class name you use when using the is keyword is a gdscript object

cullumi: Yeah you might be able to get away with a simple "script is GaeaNodeResource" check instead of the while loop? Not 100% on that, though. At the end of the day, if it works, it works.

No idea if get_instance_base_type() would be another alternative?

Zehir: I tried and he return 'Resource'

Zehir: Could not find other way to do it except that instantiate it and check the type on the dummy instance.
The other idea I had it to check for some well known method or signal names but seems quite sketchy

BenjaTK: Yeah no this is fine then. If we find another way we can change it later

PR #374, opened 2025-05-05
Add slider when number argument is min/maxed

BenjaTK:
Peek 2025-05-04 21-34

Only shows up when it has both a min and a max.

PR #384, opened 2025-05-06
Enable select all on focus for text editor fields

### LINKED ISSUE #376: Select the argument "text" content on focus
When editing multiple values in bulk you manually need to select the exiting content before adding the new value. In all other cases like this in Godot when you focus the cell the text content is selected

Image

Just select the text on focus so when you enter number they get remplaced

NA

NA

**Additional context**
NA

Zehir:
## Description

Closes #376

Enable built-in option to select the text on field focus like the Shader Graph Editor does


Zehir:
Did not think about it before but this may collide with #374

BenjaTK:
LGTM

PR #386, opened 2025-05-07
Avoid preloading graph-related resources

Zehir:
## Description

This PR replaces the **preloading** of graph-related resources with **loading** to prevent them from being loaded at runtime.

For context and discussion, see:  
https://github.com/gaea-godot/gaea/discussions/225#discussioncomment-13054708


BenjaTK:
I mean I don't see why not do this

Zehir:
Yes, the problem was that the preload do load the resource even if they don't need it and that resource was using Editor specific class that he was not able to get during runtime. And the load method load it only when called (or get from cache) so he will never load during runtime

PR #387, opened 2025-05-07
Change the way parameter names are edited

BenjaTK:
This PR changes the way parameter names are edited to disallow invalid names, improve clarity on why a name is being rejected and make accidentally changing the name harder.

It also forces parameter names to be valid identifiers (can't start with a digit; and can only have letters, digits and _). Previously existing names won't change but future ones will _have_ to be valid identifiers.

Finally, it makes parameter nodes adjust to the size of their name label.

Peek 2025-05-07 19-43

Here are the errors I got when trying to submit the invalid (red) names:
image


Zehir (approved):
LGTM

PR #334, opened 2025-04-20
Change usages of 'Axis' to 'Vector3i.Axis'

BenjaTK:
Cherry picked from #326

PR #390, opened 2025-05-10
Change usages of `Axis` with `Vector3i.Axis`

BenjaTK:
Supersedes #334

### DIFF
--- a/addons/gaea/graph/graph_nodes/generic_classes/set_operation.gd
+++ b/addons/gaea/graph/graph_nodes/generic_classes/set_operation.gd
@@ -97,9 +97,9 @@ func _get_data(output_port: StringName, area: AABB, graph: GaeaGraph) -> Diction
 
 	match get_enum_selection(0):
 		Operation.UNION:
-			for x in _get_axis_range(Axis.X, area):
-				for y in _get_axis_range(Axis.Y, area):
-					for z in _get_axis_range(Axis.Z, area):
+			for x in _get_axis_range(Vector3i.AXIS_X, area):
+				for y in _get_axis_range(Vector3i.AXIS_Y, area):
+					for z in _get_axis_range(Vector3i.AXIS_Z, area):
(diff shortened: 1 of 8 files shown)


Zehir (commented):
I dont know if we should `Vector3i.Axis.AXIS_X` instead of `Vector3i.AXIS_X` because in Godot Script editor you can ctrl click the first one but not the second one.
Also the type is `Vector3i.Axis` so I think it's expected to use `Vector3i.Axis.AXIS_X` but `Vector3i.AXIS_X` works too.

Maybe `Vector3i.AXIS_X` will be deprecated a some point

BenjaTK:
> I dont know if we should `Vector3i.Axis.AXIS_X` instead of `Vector3i.AXIS_X` because in Godot Script editor you can ctrl click the first one but not the second one.
> Also the type is `Vector3i.Axis` so I think it's expected to use `Vector3i.Axis.AXIS_X` but `Vector3i.AXIS_X` works too.
> 
> Maybe `Vector3i.AXIS_X` will be deprecated a some point

I mean if it's deprecated we can change it later but `Vector3i.Axis.AXIS_X` feels too long to me. Plus I'm pretty sure `AXIS_X` is directly referencing the enum.

Zehir:
> > I dont know if we should `Vector3i.Axis.AXIS_X` instead of `Vector3i.AXIS_X` because in Godot Script editor you can ctrl click the first one but not the second one.
> > Also the type is `Vector3i.Axis` so I think it's expected to use `Vector3i.Axis.AXIS_X` but `Vector3i.AXIS_X` works too.
> > Maybe `Vector3i.AXIS_X` will be deprecated a some point
> 
> I mean if it's deprecated we can change it later but `Vector3i.Axis.AXIS_X` feels too long to me. Plus I'm pretty sure `AXIS_X` is directly referencing the enum.

ok np, you can merge it then

PR #392, opened 2025-05-10
Fix `NumberArgumentEditor` setting min and max incorrectly

BenjaTK:
This would cause number arguments to have wrong default values.
image
(Left, before this fix. Right, after this fix)


Zehir (approved):
LGTM

PR #394, opened 2025-05-11
Abstract frames into a `GaeaGraphFrame` class

### LINKED ISSUE #388: Can't resize frame
**Describe the bug**
When the option Auto Shrink is disabled we can't resize the frame manually.

Since we can't resize the nodes anymore should we do the same with frames ?

BenjaTK:
Also fixes #388


Zehir (approved):
LGTM

PR #400, opened 2025-05-12
Add `GaeaNodeResource` tests

BenjaTK:
Adds a few tests to the nodes in the `root/` directory:
- That `_get_argument_list` is overriden.
- That its title isn't `"Unnamed"`
- That its type is in `GaeaValue.Type` and that it isn't null.
- That its arguments' types match that same condition.
- That its outputs' types match that same condition.
- That it has outputs.
- That its `get_scene` return is a valid `PackedScene`

TO-DO:
- [x] Add non-null type check
- [x] Add non-null argument type check
- [x] Add non-empty description check
- [x] Check that it has outputs
- [x] Check non-null output type
- [x] Check that `get_scene` is a valid scene.


testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
...
+## Tests that no `GaeaNodeResource`s in the root are unnamed.
+func test_are_untitled() -> void:
+	for node in nodes_in_root:
+		await assert_str(node.get_title()).is_not_equal("Unnamed")\
+			.override_failure_message("Node at %s is unnamed" % node.get_script().resource_path)
+
+## Tests that all nodes have outputs.
+func test_has_outputs() -> void:
+	for node in nodes_in_root:
+		await func(): assert_array(node.get_output_ports_list())\
```
Zehir: Godot warning "await" keyword not needed in this case, because the expression isn't a coroutine nor a signal.

testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
...
+	for node in nodes_in_root:
+		await assert_str(node.get_description())\
+			.is_not_empty()\
+			.override_failure_message("Description of node at %s is empty" % node.get_script().resource_path)
+
+
+## Check that all nodes have a valid scene.
+func test_node_scene() -> void:
+	for node in nodes_in_root:
+		await assert_object(node.get_scene())\
```
Zehir, line 125: Godot warning "await" keyword not needed in this case, because the expression isn't a coroutine nor a signal.

testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
...
+var nodes_in_root: Array[GaeaNodeResource]
+
+
+func before() -> void:
+	nodes_in_root = _get_nodes_in_folder("res://addons/gaea/graph/graph_nodes/root/")
+
+
+func _get_nodes_in_folder(folder_path: String) -> Array[GaeaNodeResource]:
+	var dir := DirAccess.open(folder_path)
+	var array: Array[GaeaNodeResource]
```
Zehir: You did not initialize the array before adding element with array.append_array, is that a godot thing ?

testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
...
+
+## Tests that no `GaeaNodeResource`s in the root push the `_get_arguments_list` warning.
+func test_is_arguments_list_overriden() -> void:
+	for node in nodes_in_root:
+		await assert_failure_await(func(): assert_error(node.get_arguments_list)\
+			.is_push_warning(("_get_arguments_list wasn't overridden in %s, node will have no arguments." % node.get_script().resource_path))
+			)
+
+## Tests that no `GaeaNodeResource`s in the root are unnamed.
+func test_are_untitled() -> void:
```
Zehir: I tried to get a warning with that but the error message was not really helpfull, can we do something about it ?

image

testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
+extends GdUnitTestSuite
+
+
+var nodes_in_root: Array[GaeaNodeResource]
+
+
+func before() -> void:
+	nodes_in_root = _get_nodes_in_folder("res://addons/gaea/graph/graph_nodes/root/")
```
Zehir: Maybe check if the script is [compiling](https://docs.godotengine.org/en/stable/classes/class_script.html#class-script-method-can-instantiate) correctly and [is a tool](https://docs.godotengine.org/en/stable/classes/class_script.html#class-script-method-is-tool).

testing/graph_nodes/general/root_nodes_testing.gd
```diff
@@ -0,0 +1,98 @@
...
+func test_has_outputs() -> void:
+	for node in nodes_in_root:
+		await func(): assert_array(node.get_output_ports_list())\
+			.is_not_empty()
+
+
+## Tests that no `GaeaNodeResource`s in the root have an invalid or null type.
+func test_null_type() -> void:
+	for node in nodes_in_root:
+		await assert_int(node.get_type())\
```
Zehir, line 91: Here too the error message is not helping
image

Zehir: I think it's for all tests, I won't add more example but it's a global problem

Zehir (changes requested):
I think all tests are needed but there is some improvements on errors message if that possible

Zehir:
I press the button to run all tests and he updated the file gdunitrunner.cfg

```json
{"included":{"res://testing":[]},"server_port":31002,"skipped":{},"version":"1.0"}
```

I also removed the file and press the run all tests button and he was able to find all tests.

I am still thinking this file should be gitignored because you define the path inside the yml file for the workflow it's not needed

testing/graph_nodes/general/root_nodes_testing.gd
BenjaTK: Yeah I don't understand why it doesn't work

Zehir: I did some tests and you need to call override_failure_message function before the comparaison. And you can also use append_failure_message function for the node path;

```gdscript
		for argument in node.get_arguments_list():
			assert_int(node.get_argument_type(argument))\

				.override_failure_message("Type of argument [b]%s[/b] is invalid" % argument)\
				.append_failure_message("Node at [b]%s[/b]" % _get_node_path(node))\
				.is_in(GaeaValue.Type.values())\

				.override_failure_message("Type of argument [b]%s[/b] of node at [b]%s[/b] is null" % [argument, _get_node_path(node)])\
				.is_not_equal(GaeaValue.Type.NULL)
```
image

The only problem is that the typing it lost after override_failure_message so you will make sure you use the correct functions.


I made a helper method to show smaller path for the node script;
```gdscript
var nodes_in_root: Array[GaeaNodeResource]
const node_root_path := "res://addons/gaea/graph/graph_nodes/root/"


func before() -> void:
	nodes_in_root = _get_nodes_in_folder(node_root_path)


func _get_node_path(node: GaeaNodeResource) -> String:
	return node.get_script().resource_path.replace(node_root_path, "")
```

But maybe we should switch to [Gut](https://github.com/bitwes/Gut). I don't know what is the best addon for that. I quite like GDUnit4 better but Gut have more experience

BenjaTK: Why would we switch to Gut? You figured out the issue and GDUnit4 is working

Zehir: I don't know, it's just an idea. I neved used both of them

BenjaTK:
> I press the button to run all tests and he updated the file gdunitrunner.cfg
> 
> ```json
> {"included":{"res://testing":[]},"server_port":31002,"skipped":{},"version":"1.0"}
> ```
> 
> I also removed the file and press the run all tests button and he was able to find all tests.
> 
> I am still thinking this file should be gitignored because you define the path inside the yml file for the workflow it's not needed

gitignored it

testing/graph_nodes/general/root_nodes_testing.gd
BenjaTK: Added these.

Zehir (commented):
LGTM

PR #402, opened 2025-05-15
Add Operation node checks and fix `GaeaNodeDataOp`

BenjaTK:
This PR adds tests for all nodes of the operation kind (`GaeaNodeFloatOp`, `GaeaNodeIntOp`, `GaeaNodeDataOp` and `GaeaNodeDatasOp`). It also fixes a problem with the `GaeaNodeDataOp` where smoothstep wasn't working correctly and it adds the power operation to it.


cullumi (approved):
Looks sane to me, and I'm assuming the tests would have caught any errors in the operation you fixed.

PR #403, opened 2025-05-17
Add the `ToHeight2D/3D` node, used to create side-view or 3D terrain.

BenjaTK:
The `ToHeight` node takes a reference data and transforms it into a 'heightmap'.

| Graph | Terrain |
| --- | --- |
| image |  image |
| image | image | 

TO-DO:
- [x] Possibly make the values in the produced data grid a gradient that goes from 1.0 at the top to the negative values.
- [x] Add tests.

### DIFF
--- a/addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
+++ b/addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
@@ -1,14 +1,20 @@
 @tool
 class_name GaeaNodeToHeight
 extends GaeaNodeResource
-## Node description.
-
+## Transforms [param reference_data] into a new data grid where the height of each column is determined by [param height_offset] + ([param reference_data] * [param displacement_intensity])
+##
+## For each cell in [param reference_data]'s [param reference_y] row, it'll get the [code]float[/code] value,
+## multiply it by [param displacement_intensity] and add [param height_offset] to it. This will be
+## the column's height, and every cell below that height (inclusive) will be full while every cell above
+## will be empty.[br][br]
+## This functions to create a heightmap, which can be used to create 2D side-view or
+## 3D terrain.[br][br]
+## [b]Note: Keep in mind the y axis in Godot is negative for up in 2D and down in 3D.[/b]
 
 enum Type {
-	TYPE_2D, TYPE_3D
+	TYPE_2D, ## Referenced data will only take into account the x coordinate of the cell.
+	TYPE_3D ## Referenced data will take into account both the x and the z coordinates of the cell.
 }
-
-var type: Type
 
 
 func _get_title() -> String:
@@ -20,9 +26,9 @@ func _get_description() -> String:
 	[param height_offset] + ([param reference_data] * [param displacement_intensity]).\n"
 	match get_enum_selection(0):
 		Type.TYPE_2D:
-			desc += "\nReferences all the x values of the [param reference_y] column."
+			desc += "\nReferences all the x values of the [param reference_y] row."
 		Type.TYPE_3D:
-			desc += "\nReferences all the x,z values of the [param reference_y] column."
+			desc += "\nReferences all the x,z values of the [param reference_y] row."
 	return desc
 
 
@@ -83,7 +89,7 @@ func _get_data(_output_port: StringName, area: AABB, graph: GaeaGraph) -> Varian
 	var height_offset: int = _get_arg(&"height_offset", area, graph)
 	var displacement: int = _get_arg(&"displacement_intensity", area, graph)
 	var data: Dictionary[Vector3i, float] = {}
-	var type: Type = get_enum_selection(0)
+	var type: Type = get_enum_selection(0) as Type
 
 	for x in _get_axis_range(Vector3i.AXIS_X, area):
 		if not reference_data.has(Vector3i(x, row, 0)):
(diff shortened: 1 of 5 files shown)


addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
```diff
@@ -0,0 +1,119 @@
...
+func _get_enum_options(_enum_idx: int) -> Dictionary:
+	return Type
+
+
+func _get_enum_option_display_name(_enum_idx: int, option_value: int) -> String:
+	return Type.find_key(option_value).trim_prefix("TYPE_")
+
+
+
+# List of all the arguments, preferably in &"snake_case".
```
Zehir: I don't think we need a comment here

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
```diff
@@ -0,0 +1,119 @@
...
+## For each cell in [param reference_data]'s [param reference_y] row, it'll get the [code]float[/code] value,
+## multiply it by [param displacement_intensity] and add [param height_offset] to it. This will be
+## the column's height, and every cell below that height (inclusive) will be full while every cell above
+## will be empty.[br][br]
+## This functions to create a heightmap, which can be used to create 2D side-view or
+## 3D terrain.[br][br]
+## [b]Note: Keep in mind the y axis in Godot is negative for up in 2D and down in 3D.[/b]
+
+enum Type {
+	TYPE_2D, ## Referenced data will only take into account the x coordinate of the cell.
```
Zehir, line 15: Could we have an axis option like the floor walker for top down heightmap instead of side view heightmap?

Like for generating islands or we could just use the fall off here instead

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
```diff
@@ -0,0 +1,119 @@
...
+func _get_argument_default_value(arg_name: StringName) -> Variant:
+	match arg_name:
+		&"displacement_intensity": return 16
+		&"gradient_intensity": return 1.0
+	return super(arg_name)
+
+
+
+
+# List of all the outputs, preferably in &"snake_case"
```
Zehir: This one too

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
```diff
@@ -0,0 +1,119 @@
...
+	var row: int = _get_arg(&"reference_y", area, graph)
+	var height_offset: int = _get_arg(&"height_offset", area, graph)
+	var displacement: int = _get_arg(&"displacement_intensity", area, graph)
+	var gradient_intensity: float = _get_arg(&"gradient_intensity", area, graph)
+	var data: Dictionary[Vector3i, float] = {}
+	var type: Type = get_enum_selection(0) as Type
+
+	var remap_offset: float = 0.0
+	if not is_zero_approx(gradient_intensity):
+		remap_offset = 64.0 / gradient_intensity
```
Zehir: What is that magic 64 value ? Should be the cell size instead or the word size ?

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
```diff
@@ -0,0 +1,119 @@
...
+
+	for x in _get_axis_range(Vector3i.AXIS_X, area):
+		if not reference_data.has(Vector3i(x, row, 0)):
+			continue
+		var z_range: Array = [0] if (type == Type.TYPE_2D) else (_get_axis_range(Vector3i.AXIS_Z, area))
```
Zehir: For performance reason we may want to move the 2D 3D check outside the axis range loop to avoid a lot of unnecessary comparisons?

testing/graph_nodes/to_height/to_height_2d_and_3d.gd
```diff
@@ -0,0 +1,34 @@
...
+
+var reference_data: Dictionary = {}
+var node: GaeaNodeToHeight
+
+
+func before() -> void:
+	var noise: FastNoiseLite = FastNoiseLite.new()
+	for x in range(AREA.position.x, AREA.end.x):
+		for z in range(AREA.position.z, AREA.end.z):
+			reference_data[Vector3i(x, 0, z)] = noise.get_noise_3d(x, 0, z)
```
Zehir: Maybe set a reference value not equal to 0 to also test the reference value

Zehir (changes requested):
Did this review outside of godot, did not have the ability to run godot for now

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
BenjaTK: That's what just the noise is for, this isn't needed

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
BenjaTK: Just an arbitrary value for the gradient, it's what worked best. The user just sees the `gradient_strength` value

addons/gaea/graph/graph_nodes/root/data/transformation/to_height.gd
BenjaTK: Done

Zehir (approved):
LGTM

PR #410, opened 2025-05-26
Update _get_data function signatures to specify return type

### LINKED ISSUE #409: Border2D throws Error when passed output from a Filter graph node
**Describe the bug**
When passing the output of a Filter graph node to a Border2D graph node, I get the following error:

> ERROR: res://addons/gaea/graph/graph_nodes/root/data/border/border.gd:58 - Trying to assign a dictionary of type "Dictionary" to a variable of type "Dictionary[Vector3i, float]".

 Tested with ThresholdFilter and DistanceFilter.

Does _not_ occur passing data directly from a Fill or FalloffMap node.

**To Reproduce**
1. In the graph panel generate a graph similar to this:
Image
2. Click Generate
3. Observer the error message in the Output tab.

**Expected behavior**
- Since the output and input channels both appear to be Data types, I expect them to be compatible
- The output should be the borders of whatever tiles are output by the filter.

**Godot Version**
4.4.1.stable.mono

**Gaea Version**
2.0

Zehir:
# Update `_get_data` Function Signatures

## Summary

This PR updates the `_get_data` function signatures to explicitly specify their return types as either:

* `Dictionary[Vector3i, float]`
* `Dictionary[Vector3i, GaeaMaterial]`

Closes #409

## Notes

Some nodes still return variant type but it's expected because he could return multiples types like the vector OP node.

### DIFF
--- a/addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
+++ b/addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
@@ -34,15 +34,15 @@ func _get_output_port_type(_output_name: StringName) -> GaeaValue.Type:
 	return GaeaValue.Type.DATA
 
 
-func _get_data(output_port: StringName, area: AABB, graph: GaeaGraph) -> Dictionary:
+func _get_data(output_port: StringName, area: AABB, graph: GaeaGraph) -> Dictionary[String, float]:
 	var texture: Texture = _get_arg(&"texture", area, graph)
 	if not is_instance_valid(texture):
 		return {}
 
-	var r_grid: Dictionary
-	var g_grid: Dictionary
-	var b_grid: Dictionary
-	var alpha_grid: Dictionary
+	var r_grid: Dictionary[String, float]
+	var g_grid: Dictionary[String, float]
+	var b_grid: Dictionary[String, float]
+	var alpha_grid: Dictionary[String, float]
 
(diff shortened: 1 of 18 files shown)


addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
```diff
@@ -34,15 +34,15 @@ func _get_output_port_type(_output_name: StringName) -> GaeaValue.Type:
...
+func _get_data(output_port: StringName, area: AABB, graph: GaeaGraph) -> Dictionary[String, float]:
 	var texture: Texture = _get_arg(&"texture", area, graph)
 	if not is_instance_valid(texture):
 		return {}
 
-	var r_grid: Dictionary
-	var g_grid: Dictionary
-	var b_grid: Dictionary
-	var alpha_grid: Dictionary
+	var r_grid: Dictionary[String, float]
```
BenjaTK: Why is this String?

BenjaTK (changes requested):

addons/gaea/graph/graph_nodes/root/resources/samplers/texture_sampler.gd
Zehir: Because of a bad copy past from the compose_range.gd ;)
Fixed in latest commit

PR #414, opened 2025-06-08
Add support for TileMap pattern rendering

Zehir:
# Add support for TileMap pattern rendering

## Summary
This PR adds support for a new `Type` in `TileMapGaeaMaterial`, enabling tiles to be rendered using TileMap **patterns** defined in the TileSet.

The Pattern material have 2 options;
- pattern_index; Pattern index in the pattern list of the [TileSet].
- pattern_offset; Pattern offset use to shift the tiles from the origin (Top left corner of the pattern).
The offset is a quick ajustement for the pattern placement because by default the patten will be placed from the top left of the pattern.

Also I placed pattern after everything because he will update multiples cell at once.

I did not update the demo scene but I tried it and works correctly

image
image


cullumi (approved):
Looks to get the job done, thanks for adding this to 2.0 @Zehir !

Issue #416, opened 2025-06-16
Parse Error: Identifier "EditorInterface" not declared in the current scope when build


gumpdev:
**Describe the bug**
I built my game and run, but the gaea generator don't work, i debug on the console and returned this error:
`Parse Error: Identifier "EditorInterface" not declared in the current scope. `
in some Gaea files

i fixed in my project, i just replace all EditorInterface to 
```
var editor_interface = Engine.get_singleton("EditorInterface")
drop_button.texture_normal = editor_interface.get_base_control().get_theme_icon(&"GuiOptionArrow", &"EditorIcons")
```
and removed some editor types on the code

**To Reproduce**
1. Use a Gaea Generator Graph
2. Build the game
3. Run the build of the game
4. See error on console

**Expected behavior**
Should work the Gaea Generator

**Godot Version**
4.4.1

**Gaea Version**
v2.0.0-beta4

aromancev:
Same here. Sounds like a critical bug to me since it's impossible to release a game or test multiplayer.

gumpdev:
@aromancev 
 
If helps you, you can work around, build your game and run in the console, track all error files and replace the `EditorInterface` (and others Godot Editor interface reference) for 
```
var editor_interface = Engine.get_singleton("EditorInterface")
editor_interface.(...)
```

gumpdev:
i created a PR #417, with the fix i made on my project

aromancev:
@gumpdev that's awesome, thanks for sharing your changes!

gumpdev:
Well, my PR was merged so i will close this issue

PR #417, opened 2025-06-22
fix export build with gaea 2.0

gumpdev:
When was build my game with Gaea 2.0 was throwing the error:
`Parse Error: Identifier "EditorInterface" not declared in the current scope when build #416`

I figure out that Godot Editor References in runtime when exported don't exists, so after i search and found this [solution](https://forum.godotengine.org/t/how-to-strip-editor-specific-code-from-build-in-godot-4-2/41978/4), was recommended to replace the EditorInterface for a get_singleton.

With this we can export game with Gaea, but we lose some autocompletes and types


cullumi (approved):
This is looking great, thanks for fixing this @gumpdev ! 😄

PR #425, opened 2025-07-26
2.0 Demos Module Integration

cullumi:
Changed 2.0's demos folder into a module connecting to the [gaea-demos 2.0 branch](https://github.com/gaea-godot/gaea-demos/tree/2.0).

For the sake of releases this shouldn't have any negative consequences. However, users cloning the repository would need to [initialize submodule files for use on their local machine](https://github.com/gaea-godot/gaea-demos/blob/main/How%20to%20make%20Git%20Modules.txt). Additionally, we should be careful not to make anything in the core plugin dependent on anything in the demos module (though we likely already do this sufficiently when testing releases).


cullumi:
Just waiting on the ability to properly duplicate GaeaGraph resource files in #431 (so that demos and tests can be decoupled, since the automated tests don't have access to the gaea-demos submodule).

addons/gaea/graph/graph_nodes/root/sampling/transformation/to_height.gd
```diff
@@ -88,6 +88,7 @@ func _get_output_port_type(_output_name: StringName) -> GaeaValue.Type:
 
 
 func _get_data(_output_port: StringName, area: AABB, graph: GaeaGraph) -> Dictionary[Vector3i, float]:
+	@warning_ignore("shadowed_variable_base_class")
```
BenjaTK, line 91: What's this for?

cullumi: Oh this was just a warning ignore added because this reference local variable shadows something from the base class. We could change the name of it instead if that's preferable. For now, I'll leave it since it's syntactically valid.

PR #426, opened 2025-07-26
1.X Demos Module Integration

cullumi:
Changed 1.X's scenes folder into a module connecting to the [gaea-demos 1.X branch](https://github.com/gaea-godot/gaea-demos/tree/1.X).

For the sake of releases this shouldn't have any negative consequences. However, users cloning the repository would need to [initialize submodule files for use on their local machine](https://github.com/gaea-godot/gaea-demos/blob/main/How%20to%20make%20Git%20Modules.txt). Additionally, we should be careful not to make anything in the core plugin dependent on anything in the demos module (though we likely already do this sufficiently when testing releases).


BenjaTK:
This looks goods to me, is it ready to be merged?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "demos"]
+	path = scenes
```
BenjaTK, line 2: Could it be renamed to demos?

PR #443, opened 2025-09-28
Change the range argument editor

BenjaTK:
This PR updates the range argument editor's slider to use the code at [detomon's assorted controls](https://github.com/detomon/godot-assorted-controls/blob/master/addons/detomon.assorted-controls/gui/range_slider/range_slider.gd). This means the new editor looks and works nicer.

https://github.com/user-attachments/assets/9a7ada30-6130-4c8d-9392-440644aa93b2


BenjaTK:
Also, I tried implementing a value/spread mode like the `ParticleProcessMaterial` editor but I couldn't get the spinboxes steps to work correctly, might try again later though

image
(what I mean)

BenjaTK:
Oops this has some unnecessary changes, will fix later

Zehir:
The range slider have some wierd beavior when you have a min value below 0 and/or a max value above 1. I am not sure what is the best way to handle it.
I also like more the semi circle than the triangle for the edges since the range icon is a donut. Maybe we could have a half donut here ?
Side question, do the range value should be clamp between 0 and 1 and for other case use a vector2i ?

BenjaTK:
> The range slider have some wierd beavior when you have a min value below 0 and/or a max value above 1. I am not sure what is the best way to handle it.
> I also like more the semi circle than the triangle for the edges since the range icon is a donut. Maybe we could have a half donut here ?
> Side question, do the range value should be clamp between 0 and 1 and for other case use a vector2i ?

I haven't noticed any weird behavior, what have you found?

The triangles match the style of the sliders in ParticleProcessMaterial. Imo it looks better that way, but I guess we can have a 3rd opinion (cough cough @cullumi )

I think Vector2i for range values would be confusing, and doing away with being able to have different min and maxes would make range values really niche.

cullumi:
I think @Zehir's observation about values outside the range holds water, but mainly in that the range it covers should be larger for certain nodes. For example, a reasonable upper range for the DistanceFilter might be 100, rather than 1. Otherwise this behavior is pretty standard for range values.

I really like the new aesthetic for the slider, and the way you can slide the whole selection around is really convenient. The triangles are a good fit, I think. Matches the arrows for the sliders, and I don't really think input bits necessarily need to match the connection point, personally.

BenjaTK:
I've changed the behavior a bit. It now matches the slider in Godot. It basically does this:

With the slider, you can't go outside the set bounds. So for example, if it's locked between 1 and 5, moving the slider will only go between those 2 numbers. When you change one of the Min and Man spinboxes to a value outside that range (and if `allow_lesser` or `allow_greater` are true), it'll adjust the slider's bounds. I can send a gif in a bit.

BenjaTK:
Peek 2025-10-13 10-57

cullumi:
> I've changed the behavior a bit. It now matches the slider in Godot. It basically does this:
> 
> With the slider, you can't go outside the set bounds. So for example, if it's locked between 1 and 5, moving the slider will only go between those 2 numbers. When you change one of the Min and Man spinboxes to a value outside that range (and if `allow_lesser` or `allow_greater` are true), it'll adjust the slider's bounds. I can send a gif in a bit.

That's good stuff, love to see it expanding the slider, as well.

PR #450, opened 2025-10-05
Make caching take in the value, not the results Dictionary

### LINKED ISSUE #432: TextureSampler cant output multiple data
**Describe the bug**
The TextureSampler node can't sample multiple colors at the same time

**To Reproduce**

Image

**Expected behavior**

**Screenshots**

Image
But with only one node

**Godot Version**
4.5

**Gaea Version**
Latest commit

**Additional context**
None

Zehir:
When loading data from the cache you still need to return the dictionary with value and type properties.

Fix #432

### DIFF
--- a/addons/gaea/graph/graph_nodes/node_resource.gd
+++ b/addons/gaea/graph/graph_nodes/node_resource.gd
@@ -59,7 +59,7 @@ var rng: RandomNumberGenerator
 ## Id in the [GaeaGraph] save data.
 var id: int = 0
 ## If empty, [method _get_title] will be used instead.
-var tree_name_override: String = "" : set = set_tree_name_override
+var tree_name_override: String = "": set = set_tree_name_override
 @export_storage var default_value_overrides: Dictionary[StringName, Variant]
 @export_storage var default_enum_value_overrides: Dictionary[int, int]
 
@@ -366,11 +366,6 @@ func _on_argument_value_changed(arg_name: StringName, new_value: Variant) -> voi
 	return
 
 
-
-
-
-
-
 #region Args
 ## Returns the value of the argument of [param name]. Pass in [param graph] to allow overriding with input slots.[br]
 ## [param area] is used for values of the type Data or Map. (See [enum GaeaValue.Type]).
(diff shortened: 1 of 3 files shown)


BenjaTK (approved):
LGTM, this makes sense.

BenjaTK:
Have you tried the unit tests? They're breaking for me

BenjaTK:
`_set_cached_data` already saved the whole value/type dictionary it seems.

BenjaTK:
The problem is in `texture_sampler.gd`, here:
image

BenjaTK:
I think we could change `_set_cached_data` to take in a `value` and `type` argument and change the calls in `traverse` and `texture_sampler.gd`'s `_get_data` to input those in.

cullumi:
> I think we could change `_set_cached_data` to take in a `value` and `type` argument and change the calls in `traverse` and `texture_sampler.gd`'s `_get_data` to input those in.

I agree. Having even just an overload that can add individual values rather than the entire dictionary would be good.

Zehir:
> I think we could change `_set_cached_data` to take in a `value` and `type` argument and change the calls in `traverse` and `texture_sampler.gd`'s `_get_data` to input those in.

But the data type can be fetch from the attribute name, here r,g,b.
I did not run the test, will check this later

BenjaTK:
> > I think we could change `_set_cached_data` to take in a `value` and `type` argument and change the calls in `traverse` and `texture_sampler.gd`'s `_get_data` to input those in.
> 
> But the data type can be fetch from the attribute name, here r,g,b.
> I did not run the test, will check this later

You can rebase now to the latest 2.0 commit so the tests will be run automatically when pushing.

BenjaTK:
It seems you did a merge commit instead of rebasing, I'd appreciate if you rebased instead.

Zehir:
I changed the methods to only store the data and not the dictionary with value and type.
I also make the methods public because they are used in other classes. They are child class but I think it's better to have them public.

BenjaTK (approved):
I like everything I just don't think they should be public. Outside classes shouldn't touch cached data and if they're child classes they can use private methods

Zehir:
Yes, you're right, I made them private. I hesitated when I wrote my last comment.

BenjaTK (approved):
LGTM!

Zehir:
Some code formatting slipped in on lines >500 in node_resource.gd. Do you want me to remove it, or should we keep it?

BenjaTK:
eh they're fine, they're good changes anyways that didn't need a whole PR lol. Leave them in

PR #451, opened 2025-10-05
Make every random number instance use node-specific `RandomNumberGenerator`

BenjaTK:


addons/gaea/graph/graph_nodes/node_resource.gd
```diff
@@ -638,11 +641,10 @@ func _is_point_outside_area(area: AABB, point: Vector3) -> bool:
...
 			point.x > area.end.x or point.y > area.end.y or point.z > area.end.z)
 
-func define_rng(graph: GaeaGraph) -> RandomNumberGenerator:
-	var rng = RandomNumberGenerator.new()
+func _define_rng(graph: GaeaGraph) -> void:
+	rng = RandomNumberGenerator.new()
 	rng.set_seed(graph.generator.seed + salt)
-	seed(graph.generator.seed + salt)
-	return rng
+	seed(rng.seed)
```
Zehir, line 647: Do you still need to define the global seed when we are using a defined RNG instance ?

Zehir (approved):
LGTB with one question

addons/gaea/graph/graph_nodes/node_resource.gd
BenjaTK: This is used for `shuffle` which sadly can't be done using a `RandomNumberGenerator`. For everything that _can_ be done using it, we should use `rng` tho.

Zehir: Ok thanks

PR #453, opened 2025-10-05
Update GDUnit and reactivate test workflows

BenjaTK:


Zehir (approved):
LGTM, did not check the gdunit code, I suppose somebody else took a look at it

PR #560, opened 2026-01-06
Add copy/pasting of nodes to/from clipboard in JSON

Zehir:
This PR add the ability to copy/paste nodes to string that can be shared on discord or other media.
It's based on the idea of [Factorio blueprints](https://wiki.factorio.com/Blueprint_string_format).
You just select node and click Copy to clipboard
image
And you will get a string in the clipboard like this :
```json
{"origin":"(740.0, 80.0)","nodes":{"6":[0,{"arguments":{"material":null,"reference_data":{},"rules":{"(0, -1, 0)":false,"(0, 0, 0)":true}},"position":"(740.0, 80.0)","salt":581472843,"type":0,"uid":"uid://br8gcsyc04ksj"}]},"connections":[]}
```
Then when you can paste nodes from the clipboard like usual node pasting.

I also had to remove the add node popup by default on the right click for 2 reasons;
- I need to be able to always have the context menu to past from clipboard
- It's annoying to have different beavior if you have a node selected or not or an empty copy buffer

Like currently copy/paste node, the material resource is not copied but I guess we can change that in the future

### DIFF
--- a/addons/gaea/editor/graph_editor/graph_edit.gd
+++ b/addons/gaea/editor/graph_editor/graph_edit.gd
@@ -690,10 +690,6 @@ func _copy_nodes(data: GaeaNodesCopy) -> void:
 
 
 func _paste_nodes(at_position: Vector2, data: GaeaNodesCopy = copy_buffer) -> void:
-	var serialized: String = data.serialize()
-	data = GaeaNodesCopy.deserialize(serialized)
-
-
 	for node in get_selected():
 		node.selected = false
 
--- a/addons/gaea/editor/graph_editor/graph_edit.gd
+++ b/addons/gaea/editor/graph_editor/graph_edit.gd
@@ -690,8 +690,8 @@ func _copy_nodes(data: GaeaNodesCopy) -> void:
 
 
 func _paste_nodes(at_position: Vector2, data: GaeaNodesCopy = copy_buffer) -> void:
-	var serialized: String = Marshalls.raw_to_base64(data.serialize())
-	data = GaeaNodesCopy.deserialize(Marshalls.base64_to_raw(serialized))
+	var serialized: String = data.serialize()
+	data = GaeaNodesCopy.deserialize(serialized)
 
 
 	for node in get_selected():
@@ -827,12 +827,11 @@ func _on_main_editor_visibility_changed() -> void:
 
 func _on_copy_from_clipboard_request() -> void:
 	var copy_data: GaeaNodesCopy = _get_copy_data(get_selected())
-	DisplayServer.clipboard_set(Marshalls.raw_to_base64(copy_data.serialize()))
+	DisplayServer.clipboard_set(copy_data.serialize())
 
 
 func _on_paste_from_clipboard_request() -> void:
-	var decoded: PackedByteArray = Marshalls.base64_to_raw(DisplayServer.clipboard_get())
-	var copy_data: Variant = GaeaNodesCopy.deserialize(decoded)
+	var copy_data: Variant = GaeaNodesCopy.deserialize(DisplayServer.clipboard_get())
 	if copy_data is GaeaNodesCopy:
 		_paste_nodes(local_to_grid(get_local_mouse_position()), copy_data)
 	elif copy_data is String:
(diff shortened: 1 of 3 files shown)


addons/gaea/editor/graph_editor/node_context_menu.gd
```diff
@@ -40,9 +42,14 @@ func populate(selected: Array) -> void:
 	add_item("Delete", Action.DELETE)
 	add_item("Clear Copy Buffer", Action.CLEAR_BUFFER)
 
+	add_separator()
+	add_item("Copy to clipboard", Action.COPY_TO_CLIPBOARD)
```
BenjaTK:
```suggestion
	add_item("Copy to Clipboard", Action.COPY_TO_CLIPBOARD)
```

addons/gaea/editor/graph_editor/node_context_menu.gd
```diff
@@ -40,9 +42,14 @@ func populate(selected: Array) -> void:
...
+	add_item("Paste from clipboard", Action.PASTE_FROM_CLIPBOARD)
```
BenjaTK:
```suggestion
	add_item("Paste from Clipboard", Action.PASTE_FROM_CLIPBOARD)
```

BenjaTK (changes requested):

addons/gaea/editor/graph_editor/graph_edit.gd
```diff
@@ -672,6 +680,10 @@ func _copy_nodes(data: GaeaNodesCopy) -> void:
 
 
 func _paste_nodes(at_position: Vector2, data: GaeaNodesCopy = copy_buffer) -> void:
+	var serialized: String = data.serialize()
+	data = GaeaNodesCopy.deserialize(serialized)
```
BenjaTK: Why is this being serialized and deserialized?

Zehir: Was a test to make sure it's was working, I forgot to remove it

Zehir: Removed

cullumi (dismissed):
A couple things:
- ~~I can't actually seem to get it to put anything into my clipboard.~~
- Could this be part of normal Ctrl-C / Ctrl-V behavior? I realize there'd be a slight conflict between the editor's "clipboard" and the system's clipboard, though I suspect that could be resolved.
- A "Ctrl-Shift-C / Ctrl-Shift-V" shortcut or similar might be nice, otherwise.

Zehir:
> A couple things:
> 
> * ~I can't actually seem to get it to put anything into my clipboard.~
> * Could this be part of normal Ctrl-C / Ctrl-V behavior? I realize there'd be a slight conflict between the editor's "clipboard" and the system's clipboard, though I suspect that could be resolved.
> * A "Ctrl-Shift-C / Ctrl-Shift-V" shortcut or similar might be nice, otherwise.

It could but we already talk about this on discord if I remember correctly and we don't want to mess with the clipboard with the ctrl+c and ctrl+v shorcuts. We can still add shorcuts with #562 later

BenjaTK:
I like Ctrl+Shift+V/C for the shortcuts. We can add them in #562 though.

BenjaTK (approved):
LGTM

BenjaTK:
Just saw this:
> I also had to remove the add node popup by default on the right click for 2 reasons;
and I disagree. I like how fast it is when you don't have a copy buffer, and it matches the functionality of the shader editor.

addons/gaea/editor/graph_editor/graph_edit.gd
```diff
@@ -749,10 +757,7 @@ func _on_gui_input(event: InputEvent) -> void:
 				return
 
 			var selected: Array = get_selected()
-			if selected.is_empty() and not is_instance_valid(main_editor.graph_edit.copy_buffer):
-				main_editor.popup_create_node_request.emit()
-			else:
-				main_editor.popup_node_context_menu_at_mouse_request.emit(selected)
+			main_editor.popup_node_context_menu_at_mouse_request.emit(selected)
```
BenjaTK: Please re-add this functionality.

BenjaTK (changes requested):

addons/gaea/editor/graph_editor/graph_edit.gd
Zehir: Updated

BenjaTK (approved):
LGTM

PR #571, opened 2026-02-01
Update to 4.6

BenjaTK:
Updates the project to Godot 4.6 and the 2 plugins to use the new `EditorDock`. Also updates the gdUnit workflow


project.godot
```diff
@@ -8,17 +8,23 @@
 
 config_version=5
 
+[animation]
```
Zehir: I think we don't need that compatibility flag

Zehir (changes requested):

project.godot
```diff
@@ -8,17 +8,23 @@
...
+
+compatibility/default_parent_skeleton_in_mesh_instance_3d=true
+
 [application]
```
Zehir, line 12: I don't think we need this migration flag, we don't use it

```suggestion
config_version=5

[application]

```

Zehir (changes requested):

cullumi (dismissed):
I take it we're still waiting on GD Unit to update to v6.1.0+ in the asset library? Cause I'm getting errors from it not being a 4.6 compatible version yet.

Zehir (approved):
LGTM

BenjaTK:
Going to wait for the other PRs I think, just in case.

Issue #573, opened 2026-02-08
Game export does not work


Zehir:
There is many references to Editor that get exported in the build that cause game crash
```bash
Godot Engine v4.6.1.rc1.official.7bfffc846 - https://godotengine.org
Vulkan 1.4.325 - Forward+ - Using Device #0: NVIDIA - NVIDIA GeForce RTX 3080

SCRIPT ERROR: Parse Error: Could not find type "EditorSettings" in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/editor_settings.gd:38)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:53)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:151)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:154)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:177)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:184)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:193)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:197)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:205)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
          at: GDScript::reload (res://addons/gaea/editor/graph_editor/graph_edit.gd:207)
SCRIPT ERROR: Parse Error: Identifier "EditorInterface" not declared in the current scope.
[...]

PR #574, opened 2026-02-09
Add `graph` property to GaeaNodeResource and remove the old reference passed around

Zehir:
This PR is the first step for #573

The idea here is to have a reference of the "owning" GareGraph on the GaeaNodeResource instead of passing around the reference. Because a lot of funtions inside GaeaNodeResource does use the reference and a GaeaNodeResource can have only one GareGraph.

Also simplify the reference to the graph inside the output node and parameter nodes. The methods was using the node->graph_edit->graph reference that is not defined on an exported game. Not sure how that was working before during gameplay

Note: Any custom nodes will need to remove the graph param so it's better to do that now than later


BenjaTK (approved):
LGTM

Issue #521, opened 2025-11-12
Separate the addon into sub-addons


BenjaTK:
**Is your feature request related to a problem? Please describe.**
Right now the organization of the folders is a bit bad, and it's bound to get worse. Plus, if you want to exclude unnecessary stuff from the export build (e.g. editor UI for Gaea), you have to be very careful to avoid excluding important stuff.

**Describe the solution you'd like**
As explained [here in the Godot documentation](https://docs.godotengine.org/en/stable/tutorials/plugins/editor/making_plugins.html#using-sub-plugins), you can create sub-plugins to separate logic. It changes nothing about the user experience, as they don't show up in the plugin list and just get activated automatically, but it could mean we could separate the `editor` scripts from the `runtime` scripts.

PR #576, opened 2026-02-10
Remove GraphNode reference from NodeResource

Zehir:
This PR is an other step for #573

This PR remove the `node` property from the `GaeaGraphNode` class.
It's now using instead a meta value with the key `&"_gaea_graph_node"`.


Zehir:
The tests are failing because this PR need #574 first

BenjaTK:
I think this needs to be finished before I can move on with #521

PR #577, opened 2026-02-10
Escape GaeaEditorSettings calls during runtime

Zehir:
This PR is an other part for #573 

The idea is to remove the reference of the `GaeaEditorSettings` class by loading directly it when needed.
+ Bonus optimization on the way


BenjaTK (approved):
LGTM

PR #578, opened 2026-02-10
Add export plugin

Zehir:
This PR is the last step for #573.

The idea here is to exclude editor-only resources and scripts from exports.

The main script is ignored (**GaeaEditorPlugin**, a.k.a. `gaea.gd`). This script holds many references to editor-only scripts and has no purpose in the game logic.

The following directories are ignored:
- `res://addons/gaea/editor/`
- `res://addons/gaea/graph/components/`
- `res://addons/gaea/assets/`

The following classes, child classes, or scenes whose root node has a matching script are also ignored:
- `GaeaGraphNode`
- `GaeaGraphFrame`
- `GaeaGraphNodeArgumentEditor`

Now, every file that gets shipped with the game should be able to run at runtime and exported game. As a result, no references to editor scripts are allowed.

### DIFF
--- a/addons/gaea/editor/export_plugin.gd
+++ b/addons/gaea/editor/export_plugin.gd
@@ -1,0 +1,79 @@
+@tool
+extends EditorExportPlugin
+class_name GaeaEditorExportPlugin
+
+
+var ignore_files: Array[Resource] = [
+	GaeaEditorPlugin,
+]
+
+
+var ignore_directory_containing: Array[Resource] = [
+	GaeaEditorExportPlugin, # res://addons/gaea/editor/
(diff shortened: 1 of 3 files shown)


BenjaTK:
This can be made more cleanly with https://github.com/gaea-godot/gaea/issues/521

Zehir:
> This can be made more cleanly with #521

Yes I think about that too but there is too many open PR to move that many files, we will get too many conflicts. I could handle it in a cleaner split if we merge the other PRs

BenjaTK:
I can work on it after we merge the other PRs, will get to reviewing today

PR #580, opened 2026-02-16
Upgrade all project files for 4.6

BenjaTK:
Ran the Upgrade Project Files tool that Godot provides.


Zehir:
I don't think it's was a good idea for the GDUnit4 plugin, there is a ton of runtime stuff that got saved, lot of cached stuff. I think you should revert the changed in the gdunit4 directory and keep only the changes for thoses that get updated on Godot load;
image

BenjaTK:
> [...] there is a ton of runtime stuff that got saved [...]

I'm not sure what you mean. These are all things Godot would do automatically if we even touched the files, so it makes sure those changes don't get added into future commits by making them now.

Zehir:
I think the addon is missing some "is part of edited scene" check, look what's got added in the file;
‎testing/rendering/gridmap_test_scene.tscn‎
‎addons/gdUnit4/src/ui/GdUnitConsole.tscn
‎‎addons/gdUnit4/src/ui/settings/GdUnitSettingsDialog.tscn‎

It's not a big issue I guess because those files are not in the final addon

BenjaTK:
> It's not a big issue I guess because those files are not in the final addon

yeahh, it's mainly repository clean-up.

Zehir (approved):

PR #581, opened 2026-02-16
Fix GaeaNodesCopy serialization to text

Zehir:
I am not sure what happen in #560. I think I just copy paste the wrong code when rebasing but the serialize/deserialize just does not works.

I remplaced the JSON to a other text based string. The JSON format was too complex to deserialize the many custom types Godot offer. So the quick solution here is to use Godot built-in methods.

I know it's no more a JSON but it's still easy to read/edit

It's using the `var_to_str` and `str_to_var` but does not allow Object to be parsed for security reasons. See issue `#80562` on Godot repo.

### DIFF
--- a/addons/gaea/resources/gaea_nodes_copy.gd
+++ b/addons/gaea/resources/gaea_nodes_copy.gd
@@ -74,76 +74,76 @@ func get_node_position(id: int) -> Vector2:
 
 
 func serialize() -> String:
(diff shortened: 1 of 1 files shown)


BenjaTK (approved):
LGTM

BenjaTK:
Could we add a test for this? Specifically for serialization and deserialization. Under the `Other` category.

PR #592, opened 2026-03-22
Fix Dictionary[Operation, Definition] type

Zehir:
For some reason, sometimes Godot thinks the type Dictionary[Operation, Definition] is not Dictionary[Operation, Definition]. I am not sure why, but for now we keep the typed dictionary instead of replacing it.

image

### DIFF
--- a/addons/gaea/runtime/graph_nodes/generic_classes/number_operation.gd
+++ b/addons/gaea/runtime/graph_nodes/generic_classes/number_operation.gd
@@ -41,8 +41,37 @@ class Definition:
 
 
 ## All possible operations.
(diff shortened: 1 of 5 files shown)


Zehir:
I am now getting random error, I need to make more changes to fix it. I think it's a new problem with Godot 4.6.2.rc2
image

Zehir:
I am not a big fan of the operation definition patching in childs class but until we get something better it's working fine

cullumi (approved):
Seems to primarily serve as a refactor. Part of me wonders if we can just wait for godot ~~4.7~~ 4.6.2 to be released before updating anything to account for it, but I don't necessarily see a problem with this particular implementation, regardless.

PR #593, opened 2026-03-22
Fix copy pasting of nodes

Zehir:
When pasting the node, we were still using the same reference, so only the data from the last copy was used for all nodes, resulting in all nodes being stacked and probably some other wrong beavior like the same name for parameters.

From bug report on [Discord](https://discord.com/channels/1142834520584892447/1142849575657410621/1485018770958450708)


cullumi (approved):
Appears to work just fine, and the original reporter of the bug (deviousdaemon on Discord) has tested this and confirmed it fixed the issue.

deviousdaemon:
Tested on my end, everything works fine.

PR #595, opened 2026-03-22
Rewrite Chunk Loader node

Zehir:
Closes #586

## Description
This is a complete rewrite of the Chunk Loader node.
The Chunk Loader now works with all tile layouts that can be set in the `TileMapLayer` node but also `GridMap`

Marked the PR as breaking because the `GaeaChunkLoader` node is gone.

## What changed and why
I removed the classes `ChunkLoader2D` and `ThreadedChunkLoader2D`. There is now a single node called `GaeaChunkLoader`

The actor is now its own resource, and there are two variants:
- GaeaChunkLoaderTilemapActor -> requires a reference to a TileMapLayer and a Node2D
- GaeaChunkLoaderGridmapActor -> requires a reference to a GridMap and a Node3D

I removed the `cell_size` property because it is no longer used, this is now calculated by the Actor class.

I made the method `get_position_conversion_gaea_to_tilemap` public and added its opposite, `get_position_conversion_tilemap_to_gaea` in the `TileMapGaeaRenderer`. These are used for tilemap-based maps to convert positions between the TileMapLayer and the Gaea map. (See #339 for detailed info). I have also optimized some calculations in theses methods.

I also fixed the method `_erase_area` in the `TileMapGaeaRenderer` that was not erasing the correct cells if the layout was not `Square`


Zehir:
@cullumi to review this PR you will mostlikly want to checkout also the demo PR at the same time : https://github.com/gaea-godot/gaea-demos/pull/5 and probably cherry pick #592 because you will get anoying errors that require a godot restart each time

addons/gaea/runtime/scene_nodes/chunk_loaders/threaded_chunk_loader_2d.gd
cullumi, line 1: This class can be deleted, threading is all handled in the Gaea Generator itself via the Gaea Thread Pool now.

Zehir: Removed

cullumi (approved):
Good to go!

PR #596, opened 2026-03-23
Fix frame renaming text input position

Zehir:
Closes #594


cullumi (approved):
Seems to have fixed it!

PR #598, opened 2026-03-23
Updated GDUnit4 to 6.1.2 and Fixed Test Orphans

cullumi:
Two things:
- Updated GD Unit to the latest version. Also fixes some errors from the last update.
- Added auto_free calls to fix orphans in test functions.


Zehir (dismissed):
Just addon update, nothing really to review but it's still wierd that the uid of this addon change every time

Zehir (approved):

PR #565, opened 2026-01-14
Add documentation to every GaeaNode

BenjaTK:
Adds missing documentation from #563 PR. Parallel to https://github.com/gaea-godot/rtd-docs/pull/24

Replacement for #564, which was in the wrong branch (oops)

PR #600, opened 2026-03-26
Add documentation to GaeaNode

Zehir:
Exactly the PR #565 but with resolved merge conflicts. The idea here is to add some documentation so we can start progress on the documentation side.


cullumi (approved):
Tracks w/ what it says it does, all changes are documentation-related.

PR #602, opened 2026-03-30
Falloff Map Expansion

### LINKED ISSUE #604: Falloff Map Axis Constraints
Along with the addition of 3D falloff maps, it should be possible to restrict falloff curves to only certain axes.

For example:
- By restricting the `y` axis, a vertically cylindrical falloff map could be produced w/ the Circle falloff shape.
- By restricting the `x` and `z` axes, a vertically fading cube could be produced w/ the Square falloff shape.

This should be implemented as a series of checkboxes for the Falloff Map.

As far as I can think of, there is not a clear alternative.

This is core to the functionality of the built-in Falloff Map node.

**Additional context**
N/A

cullumi:
# Falloff Map Expansion

Aims to close #591 ( plus #604, #603 and #590 by extension )

## Improvements

Adds support for the following:
- [x] Falloff Area configuration: Chunk, World and Custom (w/ position and size pins)
- [x] 3D falloff generation
  - Automatically stretches based on the provided area. So, an area z axis size of 1 would represent a 2D falloff map. A rectangular area w/ Circle as the Falloff Shape would yield an oval or capsule-like shape.
- [x] Custom falloff curves
  - Required support for curve resources.
- [x] Constant axes
  - The ability to produce falloff cylinders, rectangular prisms, sheets and the like by simply checking some x/y/z boxes.

Also adds:
- [x] GaeaCurve parameter resource
  - GaeaCurve abstract base w/ sample, sample_2d and sample_3d functions.
  - GaeaCurveScalar, GaeaCurve2D, and GaeaCurve3D
    - Includes value aggregation.
      - For example, if a GaeaCurve3D is asked to produce a scalar result with the `sample(offset)` method, it could return its sampled vector's `x`, `y`, or `z` values, or it could return the average of the three.
    - Includes value defaults.
      - For example, if a GaeaCurveScalar is asked to produce a Vector3 result with the `sample_3d(offset)` method, it returns a vector with exported `y` and `z` values.

Also fixes:
- A bug where the preview panel didn't actually use the correct world size in generation. Its generator settings' world size was set to the preview chunk size instead.

## Examples
Examples of what is now possible using these upgrades. The "noisy" edges are simply there to give a clearer sense of the ends of the falloff map itself. There is a green circle made with a Distance Filter at the center to show a common reference point.

World Area, Circle, Restrict X
image

Chunk Area, Circle, Restrict X and Z
image

Chunk Area, Rounded Square, Restrict Y
image

Custom Area, Squircle, Position (16, 0, 16), Size (32, 48, 32)
image


addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -136,27 +118,46 @@ Rate can be adjusted with [param start] and [param end]."""
...
+func _on_enum_value_changed(enum_idx: int, _option_value: int) -> void:
+	match enum_idx:
+		0: notify_argument_list_changed()
 
 
 func _get_arguments_list() -> Array[StringName]:
-	return [&"start", &"end"]
+	match get_enum_selection(0): # Faloff Area
+		FalloffArea.CUSTOM_AREA: return [&"position", &"size", &"curve", &"restrict_x", &"restrict_y", &"restrict_z", &"start", &"end"]
+		_: return [&"curve", &"restrict_x", &"restrict_y", &"restrict_z", &"start", &"end"]
```
Zehir: I think, for better readability, the added property should be in an if statement instead of having almost the same list twice

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -8,104 +8,85 @@ extends GaeaNodeResource
...
-			ratio = float(size_x) / float(size_y)
-		size_x_ajusted = size_x * ratio
-		size_y_ajusted = size_y * ratio
-		size_x_half = size_x * 0.5
-		size_y_half = size_y * 0.5
-		size_x_minus_y_half = size_x - size_y_half
-		size_y_minus_x_half = size_y - size_x_half
+		pos = area.position
+		center = area.get_center()
+		radii = area.size / 2
```
Zehir:
```suggestion
		radii = area.size * 0.5
```

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -169,21 +170,44 @@ func _get_output_port_type(_output_name: StringName) -> GaeaValue.Type:
...
+		FalloffArea.CHUNK_AREA:
+			area = pouch.area
+		FalloffArea.WORLD_AREA:
+			area = AABB(Vector3.ZERO, pouch.settings.world_size)
+		FalloffArea.CUSTOM_AREA:
+			var position: Vector3 = _get_arg(&"position", pouch)
+			var size: Vector3 = _get_arg(&"size", pouch)
+			area = AABB(position, size)
+
+	prints("area:", area, "( FaloffArea:", get_enum_selection(0), ")")
```
Zehir: Forgot to remove it ?
```suggestion
```

addons/gaea/runtime/resources/curves/data/curve_2d.gd
```diff
@@ -0,0 +1,38 @@
...
+## Default value when populating [method sample_3d] results.
+@export var z_default: float = 0
+
+
+func _sample(offset: float) -> float:
+	var result = curve.sample(0, offset)
+	match scalar_aggregation:
+		AggregationMethod.X: return result.x
+		AggregationMethod.Y: return result.y
+		AggregationMethod.AVERAGE, _: return (result.x + result.y)/2
```
Zehir:
```suggestion
		AggregationMethod.AVERAGE, _: return (result.x + result.y) * 0.5
```

addons/gaea/runtime/resources/curves/data/curve_scalar.gd
```diff
@@ -0,0 +1,25 @@
+@tool
+class_name GaeaCurveScalar
```
Zehir: I would use the name GaeaCurve1D instead
```suggestion
class_name GaeaCurve1D
```

addons/gaea/runtime/resources/curves/data/curve_1d.gd
```diff
@@ -0,0 +1,25 @@
...
+
+## The [Curve] to use for sampling.
+@export var curve: Curve
+## Default value when populating [method GaeaCurve.sample_2d] and [method GaeaCurve.sample_3d] results.
+@export var y_default: float = 0
+## Default value when populating [method GaeaCurve.sample_3d] results.
+@export var z_default: float = 0
+
+
+func _sample(offset: float) -> float:
```
Zehir, line 16: here too and everywhere it's used (did not make a comment for each one)
```suggestion
func _sample_1d(offset: float) -> float:
```

addons/gaea/runtime/resources/value.gd
```diff
@@ -29,6 +29,7 @@ enum Type {
 	RANGE = 100,
 	MATERIAL = 101, ## A [GaeaMaterial].
 	TEXTURE = 102, ## A [Texture].
+	CURVE = 103, ## A [GaeaCurve]
```
Zehir, line 32: I think it's would be easier to have 3 types, a curve 1D, 2D and 3D like vectors class. I find it confusing because the node may be expecting one of the other and you may now know what type you need.

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -136,27 +118,46 @@ Rate can be adjusted with [param start] and [param end]."""
...
 
 func _get_argument_default_value(arg_name: StringName) -> Variant:
 	match arg_name:
-		&"start":
-			return 0.5
-		&"end":
-			return 1.0
+		&"position": return Vector3.ZERO
+		&"area": return Vector3.ONE
+		&"curve": return null
```
Zehir, line 160: I think we need to provide some kind of default value

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
Zehir, line 115: Can you update the description with the new properties ?

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -8,104 +8,85 @@ extends GaeaNodeResource
 ## For lower [param end] values, the generated 'square' will be smaller.[br]
 ## Multiplying this with a [GaeaNodeSimplexSmooth]'s generation can create island-looking terrains.
```
Zehir, line 9: I think we should split FalloffMap for 2D, I was not able to get the previous beavior of the node. How do I generate this ?

Image

I am often getting a black square, not sure why; (added this node in the walker demo for a 2D env)

Image

Can you add migration for the node and change the UID of the new falloff node ?

Sometime I am getting this error;
```
  ERROR: res://addons/gaea/editor/graph/preview_texture.gd:62 - Invalid assignment of property or key 'visible' with value of type 'bool' on a base object of type 'Nil'.
```

addons/gaea/runtime/graph_nodes/root/sampling/generation/falloff_map.gd
```diff
@@ -136,27 +118,46 @@ Rate can be adjusted with [param start] and [param end]."""
...
 
 func _get_enums_count() -> int:
-	return 1
+	return 2
+
 
+func _get_enum_options(idx: int) -> Dictionary:
+	match idx:
+		0: return FalloffArea
+		1, _: return FalloffShape
```
Zehir, line 127:
```suggestion
		_: return FalloffShape
```

Zehir (changes requested):
I am quite lost with this node now it's not that easy to use. For the curve I would only use a single 1D curve, why do we need 3d curve ? I can't think of what that should look like

PR #609, opened 2026-04-29
Fix documentation toolkit missing graph reference

Zehir:
This PR the missing graph reference when the nodes are used in the documentation toolkit
Also update the template path for the graph nodes


cullumi (approved):
Looks good to me, just some guards and things like that.

PR #620, opened 2026-07-18
Fix gdUnit4 and update to 4.7

BenjaTK:
gdUnit was breaking so I did a clean install, that's why the uids are all different now. Also, .gitignore was set to include gaea_gdUnit4 but that's not what the folder is called. I'm not sure why it ignored some files but included others, but it's fixed now and the whole gdUnit addon should be in the repo.


Zehir (dismissed):
Hello, sure it's not a big problem to do a clean install.
Also I am not sure it's a good idea to keep the gdunit4 inside the repo, it's takeing space for nothing because the github action download it's own version right ?
There is also the WIP [global editor plugin](https://github.com/godotengine/godot/pull/119779) that we could use when it's get merged

BenjaTK:
Yeah actually it may be a good idea to remove it. I could make a PR later or if you want to you could make it.

Zehir:
I don't have access to my computer for now

PR #621, opened 2026-07-21
Remove gdUnit4 from the repository

BenjaTK:
As suggested in #620, I've removed gdUnit4 from the repository and from .gitignore. It's not needed anyways for the tests to run.


Zehir (approved):
LGTM

PR #625, opened 2026-07-26
Rename GaeaGrid to GaeaResult and allow different layer types

### LINKED ISSUE #624: GaeaLayer expansion
Right now, the only type a Gaea graph can generate and then output is a GaeaMap. This means to get data like positions or Samples you'd have to generate them manually.

GaeaLayers could have a `type` parameter which includes all wireable GaeaValue types (sample, float, int, map, etc.) defaulting to Map. Renderers would only care about map types, but then you'd also be able to obtain any other type.

It's just more convenient to have it be core. It'd be hard to build per-project.

BenjaTK:
Closes #624. Closes #534 

This PR renames `GaeaGrid` to `GaeaResult` and adds a `type` parameter to `GaeaLayer`.

image


addons/gaea/editor/graph/nodes/node.gd
```diff
@@ -148,8 +148,13 @@ func _add_slots() -> void:
 
 func _add_argument_editor(for_arg: StringName) -> GaeaEditorGraphNodeArgument:
 	var type: GaeaValue.Type = resource.get_argument_type(for_arg)
-	var scene: PackedScene = GaeaValue.get_editor_for_type(type)
+	var scene: PackedScene
+	if not resource.has_argument_editor(for_arg):
+		scene = preload("uid://i2nwlab8rau")
```
Zehir:
```suggestion
		scene = load("uid://i2nwlab8rau")
```
This is a editor file, you need to use load here

addons/gaea/runtime/resources/value.gd
```diff
@@ -41,6 +41,23 @@ enum Type {
 	VARIABLE_NAME = 305, ## Name for [GaeaNodeParameter]s.
 }
 
+## Filtered enum to only the wireable types for easy access in [GaeaLayer], for example.
+enum WireableType {
```
Zehir: Can you use `GaeaValue.is_wireable()` instead of making a new enum ?

Zehir (changes requested):
Can you also do a PR for the demos repo and also for the doc repo ?
Nice feature BTW

addons/gaea/runtime/resources/value.gd
BenjaTK: Not for exports, no.

addons/gaea/editor/graph/nodes/node.gd
BenjaTK: Done.

addons/gaea/runtime/resources/value.gd
Zehir: You can by using this ;
```gdscript
@tool
@icon("../../assets/layer.svg")
class_name GaeaLayer
extends Resource


## Whether this layer will output anything or be [code]null[/code].
@export var enabled: bool = true:
	set(new_value):
		if enabled != new_value:
			enabled = new_value
			emit_changed()
## The type of the value this layer will hold. [GaeaRenderer]s care only about
## [GaeaValue.Map] layers, but [GaeaResult] can hold any type of values.
@export var type: GaeaValue.Type = GaeaValue.Type.MAP:
	set(new_value):
		if type != new_value and GaeaValue.is_wireable(new_value):
			type = new_value
			emit_changed()

func _validate_property(property: Dictionary) -> void:
	if property.get("name") != "type":
		return

	var list: Array[String] = []
	var type_value: GaeaValue.Type = GaeaValue.Type.NULL
	for type_name: String in GaeaValue.Type:
		type_value = GaeaValue.Type.get(type_name)
		if GaeaValue.is_wireable(type_value):
			list.append(":".join([type_name.capitalize(), type_value]))
	property.set("hint_string", ",".join(list))
```

Maybe we can cache the result in some static variable to avoid recalculating it everytime because he won't change but it's not that long to process anyways

BenjaTK: done

Zehir (dismissed):
LGTM

Zehir (approved):
LGTM

PR #626, opened 2026-07-26
Add offset parameter to Noise nodes

### LINKED ISSUE #615: Add offset param to noise node
From the discussion below sometime you may want for 2D game to have 2 similar noises, the idea here is to edit the noise node to add an offset parameter that would be a vector to offset the generation. The key of the returned data won't be affected by this

From : https://github.com/gaea-godot/gaea/discussions/613

BenjaTK:
Closes #615 

image
image


Zehir (approved):
What a quick feature request, LGTM

